PR #130, opened 2019-09-18
[T] Update the services

ilyasmez:
Sometimes we want to have a custom offset instead of `window.innerHeight` in `isScrolledPastElBottom`.

And on the way, I deleted the obsolete `loadJSAsync` function, we should use `loadScript` instead :)


src/services/utils.js
```diff
@@ -78,18 +78,6 @@ export const getRandomInt = (min, max) => Math.floor(Math.random() * (max - min
 
 export const circleToPath = (cx, cy, r) => `M ${cx} ${cy} m -${r}, 0 a ${r},${r} 0 1,0 ${r * 2},0 a ${r},${r} 0 1,0 -${r * 2},0`;
 
-export const loadJSAsync = (e, n, o) => {
```
SinisaG: This is a breaking change possibly. Not sure anyone else is using it atm though.

OriginalEXE: I don't think anyone is, but I still think we should follow the SemVer when we release this and bump the major version.

PR #135, opened 2019-09-24
[T] Upgrade dependencies

OriginalEXE:
I have finally managed to upgrade the dependencies. Won't admit how much time it took... Let's just say that the changelog shown here is deceptively simple :see_no_evil: 

I had to create our own fork of `vue-cli-plugin-storybook` in order to upgrade its dependencies as well. It looks like it's not super-maintained at the moment so let's see how the situation develops.

We are also using the rc of the `@vue/cli-service`. Not ideal but it seems stable for now, and since it's used for `homeday-blocks` deployment only anyways and we are consuming the components directly, it should not be possible for it to break our production builds.

It will be nice to be on the latest version of Storybook :tada:


webpack-alias.js
```diff
@@ -1,3 +1,6 @@
+// Needed until https://github.com/benmosher/eslint-plugin-import/issues/1396
```
viniciuskneves: I would recommend us to subscribe to this issue and try it again once closed.

OriginalEXE: :+1: Did so

package.json
```diff
@@ -14,39 +14,40 @@
...
+    "eslint": "^6.4.0",
     "eslint-plugin-vue": "5.2.3",
     "lint-staged": "9.3.0",
     "node-sass": "^4.12.0",
-    "raw-loader": "3.0.0",
-    "sass-loader": "7.1.0",
-    "vue-cli-plugin-storybook": "0.6.1",
+    "raw-loader": "3.1.0",
+    "sass-loader": "8.0.0",
+    "vue-cli-plugin-storybook": "homeday-de/vue-cli-plugin-storybook#master",
```
SinisaG: Last commit was 3 months ago - https://github.com/storybookjs/vue-cli-plugin-storybook/commits/master :/

PR #139, opened 2019-10-04
[F] Close gallery overlay on esc or overlay click

OriginalEXE:
https://homeday.atlassian.net/browse/CE-845

I have removed some `pointer-events` so that in most cases the gallery overlay receives the click.

I did not touch the vue zoomer since on mobile we use the whole area for pinch and zoom as well as swiping.

### DIFF
--- a/src/components/gallery/HdGalleryOverlay.vue
+++ b/src/components/gallery/HdGalleryOverlay.vue
@@ -63,7 +63,7 @@ export default {
         this.$refs.gallery.goNext();
       } else if (['Left', 'ArrowLeft'].includes(e.key)) {
         this.$refs.gallery.goPrev();
-      } else if (e.key === 'Escape') {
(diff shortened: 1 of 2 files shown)


src/components/gallery/HdGalleryOverlay.vue
```diff
@@ -60,8 +63,18 @@ export default {
         this.$refs.gallery.goNext();
       } else if (['Left', 'ArrowLeft'].includes(e.key)) {
         this.$refs.gallery.goPrev();
+      } else if (e.key === 'Escape') {
```
ilyasmez: I believe we should also check for `Esc` (Used by IE and Edge)
```suggestion
      } else if (['Esc', 'Escape'].includes(e.key)) {
```

OriginalEXE: Nice catch, thanks!

PR #148, opened 2019-10-09
[F] Add Welcome "story"

ilyasmez:
There is not much written in the welcome page, so please consider this as a start and everybody is invited to contribute.

I also added the [Viewport addon](https://github.com/storybookjs/storybook/tree/master/addons/viewport) which I find pretty useful. We can add our custom viewports to represent our breakpoints (mobile, tablet, desktop, desktop-wide).


viniciuskneves:
image

Is it possible to close `actions/knobs` in a certain story? If so it would be nice in this first one.

viniciuskneves:
In case someone doesn't want to run the project, it looks like that:

image

ilyasmez:
@viniciuskneves
> Is it possible to close `actions/knobs` in a certain story? If so it would be nice in this first one.

Yes it's possible. I went with hiding the toolbar as well 👍  Thanks for the suggestion.

> Just wouldn't merge/deploy it until we get homeday-de/homeday-blocks back.

We are not using the npm package here, but the github repo, so it's safe to go with it that way. @SinisaG will take care of using the npm package later (we talked with Oleg).

config/storybook/addons.js
```diff
@@ -3,3 +3,4 @@ import '@storybook/addon-actions/register';
 import '@storybook/addon-links/register';
 import '@storybook/addon-notes/register';
 import '@storybook/addon-knobs/register';
+import '@storybook/addon-viewport/register';
```
SinisaG: :+1:

PR #154, opened 2019-10-11
[F] Updated breakpoints service, Responsive component

OriginalEXE:
https://homeday.atlassian.net/browse/FET-83
https://homeday.atlassian.net/browse/CE-865

This completely revamps the breakpoints service and introduces the `Responsive` component. I also added an initial structure for documenting our services.

Another breaking change is a rework of the `HdExpandText`. Previously it relied on `isMobile` form the old breakpoints service, but since we no longer have a concept of `mobile`, I made it configurable with breakpoints instead.

Open to the feedback of course.


README.md
```diff
@@ -91,3 +91,10 @@ import HdDynamicForm from 'hd-blocks/components/form/HdDynamicForm.vue';
 
 ## Customize configuration
 See [Configuration Reference](https://cli.vuejs.org/config/).
+
+## Services
```
viniciuskneves: I would add the link to the services/README here instead of listing all the services here. I would list the services as an index in the beginning of services/README file.

src/components/HdExpandText.vue
```diff
@@ -33,28 +33,18 @@
...
-};
-
 export default {
   name: 'HdExpandText',
   props: {
+    // Can be an object with keys set to breakpoints and values the desired
+    // lines count
     lines: {
-      type: Number,
+      type: [Number, Object],
```
viniciuskneves: I'm personally against multiple types. I would have 2 different properties and for the "linesBreakpoints" I would also add the validator.

README.md
OriginalEXE: Agreed, done :+1:

src/components/HdExpandText.vue
OriginalEXE: Why so? If we have two properties that basically do the same thing, how do we determine which one takes precedence?

I can add a validator for the argument, that might be a good idea (since it's a form of documentation on its own).

viniciuskneves: I don't like it because you've to do `typeof` checks in the code to guarantee different behaviors. But as I said, it is personal only.

OriginalEXE: Yeah, I can understand that. I don't think it's ideal either. However, I don't see a version with two different props as any better. The `typeof` checks would have to be replaced with checking whether one argument is set, and if not falling back to another (in essence doing the same thing, differently), but with a drawback of having an additional prop to maintain and arbitrarily deciding which one takes preference.

Hopefully, the rest of the team weighs in and we can adjust as per majority preference.

OriginalEXE: I checked how some other component libraries are doing it and some of them are also following the "multiple types" approach (for example `Button` component in Vuetify for the `href` prop supports a `string` or an `object`, `Dropdown` component inBuefy for the `can-close` prop supports `boolean` and `array`.

Not saying that proves it's the best way to do it, but we would not be setting a precedent :smile:

viniciuskneves: Hahahah, great! Nice catch! I'm already convinced about multiple types hehehehe

SinisaG: I have no strong preference here, both have ups and downs, as you've discussed. ;)

src/services/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+You can now import a helper component and render contents based on your breakpoints matching or not:
+
+```vue
+<script>
+import { Responsive } from 'hd-blocks/services/breakpoints';
+</script>
+
+<template>
+<Responsive #default="{ matches, indeterminate }">
```
SinisaG: Can you please add example of `matches` and `indeterminate`? I think is not fully clear how those object look like, even though, there is a description under.

src/services/breakpoints.js
```diff
@@ -1,21 +1,70 @@
...
-  breakpoints = { ...breakpoints, ...newBreakpoints };
+  breakpoints = newBreakpoints;
 };
 
-export default {
-  isMobile,
-  isTablet,
-  isDesktop,
-  setBreakpoints,
+export const getBreakpoints = () => breakpoints;
```
SinisaG: It looks a bit strange, having component inside of service here. And this component is also persisting state (breakpoints. right?). How about decoupling setting and getting breakpoints and keep it in `breakpoints` service and taking the component out of there. Then we'll have `Responsive` component, which could take also `breakpoints` object, if not passed, it would check the service, if there are global breakpoints? 
So you would also be able to pass specific breakpoints to the responsive component, instead of overwriting them on global level, if you need on the fly modification. Then I think the split is also clearer, where breakpoints service is responsible for setting and updating global breakpoints and just that. And Responsive component is responsible of consuming the breakpoints and falling back.

src/services/README.md
OriginalEXE: Will do :)

src/services/breakpoints.js
OriginalEXE: Thanks for a great idea, that's a good suggestion. Will implement it.

src/services/README.md
OriginalEXE: I have added the examples into the docs immediately below.

src/services/breakpoints.js
OriginalEXE: Done :)

OriginalEXE:
As per @SinisaG feedback, I have extracted the `Responsive` component from the service and into a dedicated file.

I reworked the docs and added a story for the storybook so that the component is listed there as well.

src/components/HdExpandText.vue
```diff
@@ -121,6 +121,23 @@ export default {
...
     },
+    getLinesCount() {
+      if (typeof this.lines === 'number') {
+        return this.lines;
+      }
+
+      let linesCount = this.linesFallback;
+
+      Object.keys(this.lines).forEach((breakpoint) => {
+        if (!mediaMatches(breakpoint)) {
```
SinisaG: For future improvement, it might make sense, to be able to pass breakpoints as prop here as well and then fallback to the main ones. Just to keep behavior consistent. If you pass `lines` as object 

```javascript
{
xl: 4,.
l: 3,
}
```

but don't set global endpoints, what would happen?

src/services/README.md
```diff
@@ -0,0 +1,52 @@
...
+<HdResponsive #default="{ matches, indeterminate }">
+  <div v-if="indeterminate">Matching could not be reliably determined (server side rendering, unit tests environment)</div>
+  <div v-else-if="matches.xl">xl breakpoint matched</div>
+  <div v-else-if="matches.m">m breakpoint matched</div>
+</HdResponsive>
+</template>
+```
+
+As seen in the example above, `HdResponsive` component exposes two arguments:
+- `matches`: Object with the keys matching your breakpoints object and boolean values (`true`: matches, `false`: does not match the current viewport)
```
SinisaG: Is worth mentioing breakpoints prop here? I know you did it above already. 

This looks better. :ok_hand:

OriginalEXE: Good point. I mentioned it in the docs.

src/components/HdExpandText.vue
OriginalEXE: Agreed, I will make a not to add this enhancement. Right now it would simply fall back to the `linesFallback`, but it would be nice to be able to set specific breakpoints since the component relies on them.

PR #158, opened 2019-10-15
[F] HdButton

viniciuskneves:
I've seen the `<button>` being customized in multiple places as:
- Multiple views inside Customer App
- https://github.com/homeday-de/homeday-blocks/blob/develop/src/components/form/HdDynamicForm.vue
- https://github.com/homeday-de/homeday-blocks/blob/develop/src/components/buttons/HdLoaderButton.vue

It might also be useful for sliders and Kirby as some buttons are not inside forms (as the one created by `HdDynamicForm`).

- Added some update to README to run tests for just one component
- Fixed the path of HdGallery inside Welcome component


OriginalEXE:
Do you think it's worth the effort maintaining this since we need to re-work the buttons anyway as part of the design system?

viniciuskneves:
> Do you think it's worth the effort maintaining this since we need to re-work the buttons anyway as part of the design system?

Fuck, forgot that hehehe

viniciuskneves:
Nevertheless I think at the end we gonna update some stuffs on top of it. We will still have a primary button at the end.

OriginalEXE:
I think it would make more sense to have a Button component with a `type` (or a `variation`) prop rather than a separate component for each button.

Not saying it's necessarily a bad idea to add this now (design system might take a while), just questioning it :)

viniciuskneves:
> I think it would make more sense to have a Button component with a `type` (or a `variation`) prop rather than a separate component for each button.
> 
> Not saying it's necessarily a bad idea to add this now (design system might take a while), just questioning it :)

Yeah, I thought about it as well but to be honest I don't have a clear answer in my mind. It would be flexible to have `<HdButton type="primary,secondary..."/>` but I having `<HdPrimaryButton />` is super explicit and self documented. I would like to get input from @ilyasmez @SinisaG @drapisarda @jpDenisO as well so we can decide in which direction to go.

ilyasmez:
When we briefly discussed this I thought that we will have one component for all the buttons, because like this I don't think it will bring so much value or it will stop the adjustments we have in the other projects, even the disabled prop is not necessary because the `button` element is in the root of the template.

I would go with @OriginalEXE's suggestion of having a component that takes a `type` prop, and maybe wait for the design system to be finalized to avoid wasted work.

viniciuskneves:
@OriginalEXE and @ilyasmez could you take a look once again?

@SinisaG @drapisarda @jpDenisO could you also give it a try?

src/components/buttons/HdButton.vue
```diff
@@ -0,0 +1,43 @@
+<template>
+  <button :class="computedClasses">
```
SinisaG: Did you consider exposing also some `button` attributes(https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#Attributes)? I would suggest adding at least `type` - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type as prop. 

If you want to avoid setting each one, it could also take the attributes object that then dynamically renders passed in attributes.

OriginalEXE: Don't think this is needed since the `button` is a root component, so any attributes added to the component in the template will be "rendered" on it.

viniciuskneves: We won't need to add `<button>` attributes for the reason @OriginalEXE pointed. If we pass `:disabled="disabled"` or so it will be added to the root HTML tag which is the `<button>` itself.

SinisaG: Ah true, it's root. All good. :+1:

viniciuskneves:
I'm going to use `<slot>` here as well.

PR #170, opened 2019-10-17
[T] Upgrade dependencies

OriginalEXE:
This upgrades to the newly released version 4.0.x of Vue CLI.
We can now also safely upgrade to the latest core-js version, as well as `babel-jest`, which is nice.


OriginalEXE:
Yup, exactly. I tried running a storybook, building in library mode and running unit tests, and all was ok.

We are anyways consuming the components directly so this, in theory, should not be able to break any production code, but still work being careful :)

viniciuskneves:
Let's see how many patches/minors will be released before this PR is actually merged 😆

PR #176, opened 2019-10-17
[F] <HdLink>

viniciuskneves:
Link component where we can use it instead of defining `<a>` and `<router-link>`.

It switches between `<a>` and `<router-view>` based on the `to` property which is used by `<router-link>` by default. I've not added the other properties from `<router-view>` as I dind't find any use case of them but we can add when needed.

It is an attempt to move our spread CSS rules to components approach.

### DIFF
--- a/src/stories/HdLink.stories.js
+++ b/src/stories/HdLink.stories.js
@@ -21,7 +21,7 @@ storiesOf('HdLink', module)
       <HdLink
         :text=text
         :modifier=modifier
-        href="www.homeday.de"
+        href="https://www.homeday.de"
       />
     `,
   }), {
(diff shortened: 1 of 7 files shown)


src/stories/HdLink.stories.js
```diff
@@ -0,0 +1,31 @@
...
+      modifier: {
+        type: String,
+        default: select('Modifier', ['primary', 'secondary'], 'primary'),
+      },
+    },
+    template: `
+      <HdLink
+        :text=text
+        :modifier=modifier
+        href="www.homeday.de"
```
ilyasmez: Just to make it work as an absolute url :)
```suggestion
        href="https://www.homeday.de"
```

viniciuskneves:
> Looks good overall, nice job. Any specific reason why you went with the `text` prop instead of a slot?

No, just didn't find any usage that would require `<slot>` for now.

ilyasmez:
@viniciuskneves I vote for `slots`, at least it will save us from the prop naming inconsistencies we already have (`text`, `message`, `label`...) 😄

viniciuskneves:
@ilyasmez @OriginalEXE @SinisaG could you take a look again? I've switched to slots now.

PR #177, opened 2019-10-17
[F] Range Input Component

drapisarda:
A first attempt! Let me know what you think of it!

The main idea is to have some HTML elements under a real and transparent input[type="range"] and then maintain the sync between UI and input value. 

It also works using the keyboard!


src/components/form/HdRange.vue
```diff
@@ -0,0 +1,274 @@
...
+    },
+    maxValue() {
+      this.updateRangeDecoration();
+    },
+  },
+  mounted() {
+    this.input = this.$el.querySelector('input');
+    this.progress = this.$el.querySelector('.range__progress');
+    this.progress.style.width = '1px';
+    this.thumb = this.$el.querySelector('.range__thumb');
```
ilyasmez: I think it's better to use [$refs](https://vuejs.org/v2/api/#ref) instead of `this.$el.querySelector`

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,274 @@
...
+
+      &::-ms-track,
+      &::-ms-fill-lower,
+      &::-ms-fill-upper {
+        @include track-style;
+      }
+    }
+
+    &__decoration {
+      $decoration : &;
```
ilyasmez: Doesn't seem like you're using this variable somewhere

src/stories/HdRange.stories.js
```diff
@@ -0,0 +1,45 @@
...
+      rangeStep: {
+        type: Number,
+        default: number('Step', 1),
+      },
+    },
+    methods: {
+      getValue() {
+        console.log('lol');
+      },
+    },
```
ilyasmez: This is not used, maybe you can change it to print the value:
```js
watch: {
   value(value) {
     console.log('value', value); // or using the action addon
  },
},
```

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,274 @@
...
+    updateRangeDecoration() {
+      if (this.input.value > this.maxValue) {
+        this.input.value = this.maxValue;
+      }
+
+      if (this.input.value < this.minValue) {
+        this.input.value = this.minValue;
+      }
+
+      this.displayValue = this.input.value;
```
ilyasmez: `displayValue` is not defined or used, this line can be removed.
Instead it seems like the right place to emit the `input` event:
```
this.$emit('input', parseFloat(this.input.value));
```

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    this.updateRangeDecoration();
+  },
+  computed: {
+  },
+  methods: {
+    updateRangeDecoration() {
+      if (this.input.value > this.maxValue) {
+        this.input.value = this.maxValue;
```
ilyasmez: (Same this for the if statement bellow)

Instead of assigning to `this.input.value`, just emit an event `input` with the value you need, like so:
```js
this.$emit('input', this.maxValue);
```

v-model of the component (the one set on the parent) will take care of the binding :)

src/stories/HdRange.stories.js
```diff
@@ -0,0 +1,45 @@
...
+  .addDecorator(FormWrapper)
+  .add('default 🎛', () => ({
+    components: { HdRange },
+    template: `
+      <div>
+        <HdRange
+          :minValue=minValue
+          :maxValue=maxValue
+          :rangeStep=rangeStep
+          :value="value"
```
ilyasmez: To keep the same api for the form components (necessary for HdDynamicForm), we prefer to use `v-model`.
```suggestion
          v-model="value"
```

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,274 @@
...
+      default: 0,
+    },
+    rangeStep: {
+      type: Number,
+      default: 1,
+    },
+  },
+  watch: {
+    value() {
+      this.input.value = this.value;
```
ilyasmez: This is not needed it when using `v-model`.

src/components/form/HdRange.vue
viniciuskneves: Agree

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,274 @@
...
+      this.updateRangeDecoration();
```
jpDenisO: the method `updateRangeDecoration` is already set as callback for @input and @change, wouldn't it get triggered x3 on each change?

jpDenisO: I personally would keep it only here at the watcher, and leave the events for callbacks passed from a parent component

src/components/form/HdRange.vue
drapisarda: From the docs I thought it should be used just for components, but I see it's not. I go for it

src/components/form/HdRange.vue
drapisarda: Ok. Do we need to emit this event? Doesn't it emit by itself?

drapisarda: Also, I made a method called _normalizeValue_ . It's always used before the uodateRangeDecoration but it's more descriptive.

src/components/form/HdRange.vue
drapisarda: Got it!

src/components/form/HdRange.vue
ilyasmez: Yes, we need to emit the event, so the parent can know about the changes.

src/components/form/HdRange.vue
drapisarda: I had to set it on both change and input because IE11 emits change instead of input during the drag. So I have to keep them both. Unfortunately, I can not make it work at the watcher!

src/stories/HdRange.stories.js
drapisarda: I did not found a way to do it. I have to validate that value, so I can not use v-modal in the story, otherwise I get this 
image

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,315 @@
...
+      const thumbWidthOffset = this.thumb.offsetWidth * valuePercentage / 100;
+      this.thumb.style.transform = `translateX(${valuePercentageInputWidthPixels - thumbWidthOffset}px)`;
+    },
+    handleFocus() {
+      this.isActive = true;
+    },
+    handleBlur() {
+      this.isActive = false;
+    },
+    currentValue() {
```
ilyasmez: You're using this method as a modal (`v-modal`)? I'm not sure that will work. A modal needs to be settable and gettable.
I suggest adding `currentValue` as a data property, renaming this method and calling it in the beginning of `updateRangeDecoration()`

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,315 @@
...
+      :name="name"
+      ref="input"
+      :v-model="currentValue"
+      :required="required"
+      :disabled="disabled"
+      :min="minValue"
+      :max="maxValue"
+      :step="rangeStep"
+      @input="updateRangeDecoration"
+      @change="isLegacy ? updateRangeDecoration : null"
```
ilyasmez: `v-modal` handles this (with the edges cases that you mentioned in the other comment as well).
If you make the changes as I suggested in the other comment, watching `currentValue` will be enough.

src/stories/HdRange.stories.js
drapisarda: I'm using the _v-model_ on the input element. The component still have _value_ as a prop.

src/components/form/HdRange.vue
drapisarda: I was using it to validate the value, but the range element input value does it by itself. So I remvove this.

src/components/form/HdRange.vue
drapisarda: Let's sync on this.

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,313 @@
...
+      v-model="currentValue"
+      :id="name"
+      :name="name"
+      :required="required"
+      :disabled="disabled"
+      :min="minValue"
+      :max="maxValue"
+      :step="rangeStep"
+      @input="updateRangeDecoration"
+      @change="isLegacy ? updateRangeDecoration : null"
```
viniciuskneves: I would add a test for this guy as well. As if `isLegacy` is set it won't call `updateRangeDecoration`.

viniciuskneves: You could also add it to the story so we could play with it as well.

drapisarda: Nice! I wrote the test and got that that feature was not working. I made it differently and works great. That's TDD's best flavour. Thank you, man.

I also wrote tests for the other listeners. Big Up!

viniciuskneves:
Hey @drapisarda as I mentioned in Slack, I'm going to review it after PEW, unless you really need this component for your current development then let me know!

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,320 @@
...
+
+      this.currentValue = this.currentValue - (this.currentValue % this.rangeStep);
+
+      this.$emit('input', parseFloat(this.currentValue));
+    },
+    updateRangeDecoration() {
+      this.normalizeAndEmit();
+
+      // the percentage of the value, between max and min. I.e. : 15 is the 50% between 10 and 20
+      const valuePercentage = this.currentValue / ((this.$props.maxValue - this.$props.minValue) / 100);
```
ilyasmez: The current implementation will work properly **only** if the minValue is `0`, this should fix it:
```suggestion
      const valuePercentage = (this.currentValue - this.minValue) / (this.maxValue - this.minValue) * 100;
```

And yeah, you don't have to use `$props`, (not like in some other frameworks, react... ehm) :D

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,320 @@
...
+      v-model="currentValue"
+      :id="name"
+      :name="name"
+      :required="required"
+      :disabled="disabled"
+      :min="minValue"
+      :max="maxValue"
+      :step="rangeStep"
+      @input="updateRangeDecoration"
+      @change="handleChange"
```
ilyasmez: Now that we are using `v-model`, we don't have to handle these events manually. We can remove them.

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,320 @@
...
+    handleBlur() {
+      this.isActive = false;
+    },
+    handleChange() {
+      if (!this.isLegacy) {
+        return;
+      }
+
+      this.updateRangeDecoration();
+    },
```
ilyasmez: We won't need this one as well.

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,320 @@
...
+      default: 100,
+    },
+    rangeStep: {
+      type: Number,
+      default: 1,
+    },
+    isLegacy: {
+      type: Boolean,
+      default: false,
+    },
```
ilyasmez: This one can be removed as well.

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,320 @@
...
+    maxValue() {
+      this.updateRangeDecoration();
+    },
+    rangeStep() {
+      this.updateRangeDecoration();
+    },
+    value() {
+      this.currentValue = this.value;
+      this.updateRangeDecoration();
+    },
```
ilyasmez: Here we can make use of the modal (that's how we won't need to handle the change/input events ourselves):

```js
value() {
  this.currentValue = this.value;
},
currentValue() {
  this.updateRangeDecoration();
},
```

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,320 @@
...
+    normalizeAndEmit() {
+      if (this.currentValue > this.maxValue) {
+        this.currentValue = this.maxValue;
+      } else if (this.currentValue < this.minValue) {
+        this.currentValue = this.minValue;
+      }
+
+      this.currentValue = this.currentValue - (this.currentValue % this.rangeStep);
+
+      this.$emit('input', parseFloat(this.currentValue));
```
ilyasmez: Small optional thing:
You don't have to parse the value to float yourself you can use [`v-modal.number`](https://vuejs.org/v2/guide/forms.html#number) (it does exactly the same thing, I only find it cleaner, and it's always nice to parse as early as possible).

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,320 @@
...
+      this.updateRangeDecoration();
+    },
+    value() {
+      this.currentValue = this.value;
+      this.updateRangeDecoration();
+    },
+  },
+  mounted() {
+    this.progress = this.$refs.progress;
+    this.progress.style.width = '1px';
```
SinisaG: Maybe a comment why this is needed?

src/components/form/HdRange.vue
drapisarda: Supper man

src/components/form/HdRange.vue
drapisarda: It does not work without those!

src/components/form/HdRange.vue
drapisarda: I need it to handle the change event on IE11

src/components/form/HdRange.vue
drapisarda: 🚀

tests/unit/components/form/HdRange.spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+  test('renders component', () => {
+    expect(wrapper.html()).toMatchSnapshot();
+  });
+
+  test('renders with preselected value', () => {
+    expect(parseFloat(wrapper.find('input').element.value)).toBe(initialValue);
+  });
+
+  test('manages the max value', () => {
```
SinisaG: Maybe a better wording here would be - `respects the limit of max value`

tests/unit/components/form/HdRange.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    wrapper.setProps({
+      value: maxValue * 2,
+    });
+
+    expect(parseFloat(wrapper.find('input').element.value)).toBe(maxValue);
+    expect(wrapper.emitted('input')[1][0]).toEqual(maxValue);
+  });
+
+  test('menages the min value', () => {
```
SinisaG: Similar here: `Limits the value to passed minimum value`

src/components/form/HdRange.vue
drapisarda: Very nice!

tests/unit/components/form/HdRange.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    wrapper.setProps({
+      isLegacy: false,
+    });
+
+    wrapper.find('input').trigger('change');
+
+    expect(mockedUpdateRangeDecoration).toHaveBeenCalledTimes(0);
+  });
+
+  test('input focus emit is managed', () => {
```
SinisaG: If I get it right, we're testing here that onFocus is called when `focus` event happens? I would suggest updated wording here also e.g. `focusHanlder is called on focus`

tests/unit/components/form/HdRange.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    const mockedHandleFocus = jest.fn();
+    wrapper.setMethods({ handleFocus: mockedHandleFocus });
+
+    wrapper.find('input').trigger('focus');
+
+    expect(mockedHandleFocus).toHaveBeenCalledTimes(1);
+  });
+
+  test('input blur emit is managed', () => {
```
SinisaG: Respectively here: `blurHanlder is called on blur `

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,335 @@
...
+    },
+    value() {
+      this.currentValue = this.value;
+    },
+    currentValue() {
+      this.updateRangeDecoration();
+    },
+  },
+  mounted() {
+    // the progress bar needs a starting fixed width to be then transformed in updateRangeDecoration() with scaleX property
```
SinisaG: Thanks for comment. :+1:

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,329 @@
...
+    },
+    blurHandler() {
+      this.isActive = false;
+    },
+  },
+};
+</script>
+
+<style scoped lang="scss">
+.range {
```
SinisaG: I suggest to explicitly import `mixins.scss` here, so we slowly shift away from the patter of having them injected with webpack in parent project.

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,329 @@
...
+      :max="maxValue"
+      :step="rangeStep"
+      @focus="focusHandler"
+      @blur="blurHandler"
+    />
+    <div class="range__decoration" ref="decoration">
+      <div class="range__progress" ref="progress" />
+    </div>
+    <ul v-if="displayStepBullets" class="range__steps">
+      <li v-for="(steps, i) in stepsAmount" :key="i" />
```
SinisaG: This is the "problematic" part.

tests/unit/components/form/HdRange.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    wrapper = wrapperFactory();
+  });
+
+  test('renders component', () => {
+    expect(wrapper.html()).toMatchSnapshot();
+  });
+
+  test('renders with preselected value', () => {
+    expect(parseFloat(wrapper.find('input').element.value)).toBe(initialValue);
+  });
```
viniciuskneves: The second assertion could be inside the first test as you initialize the component with a value already.

tests/unit/components/form/HdRange.spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+    wrapper.find('input').trigger('focus');
+
+    expect(mockedFocusHandler).toHaveBeenCalledTimes(1);
+  });
+
+  test('When the focus event is emitted, the input is in an active status', () => {
+    wrapper.find('input').trigger('focus');
+
+    expect(wrapper.vm.isActive).toBe(true);
```
viniciuskneves: I think those two guys can also be inside the same test (if I'm missing some context, please let me know).

tests/unit/components/form/HdRange.spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+    wrapper.find('input').trigger('blur');
+
+    expect(mockedBlurHandler).toHaveBeenCalledTimes(1);
+  });
+
+  test('When the blue event is emitted, the input is in an non active status', () => {
+    wrapper.find('input').trigger('blur');
+
+    expect(wrapper.vm.isActive).toBe(false);
```
viniciuskneves: Same here

src/components/form/HdRange.vue
```diff
@@ -0,0 +1,329 @@
...
+      default: 1,
+    },
+    value: {
+      type: Number,
+      default: 0,
+    },
+    displayStepBullets: {
+      type: Boolean,
+      default: true,
+    },
```
viniciuskneves: I would add a test to cover this property as it is not tested anywhere and apparently, as pointed out by @SinisaG, there is a bug for this guy.

src/components/form/HdRange.vue
drapisarda: I fixed it, giving the number of steps as integer division and give a relative padding-right to the unsorted list. The value of the padding-right is the proportion of module of the division on the range "size".

src/components/form/HdRange.vue
drapisarda: So, same for var?

tests/unit/components/form/HdRange.spec.js
drapisarda: 👍

tests/unit/components/form/HdRange.spec.js
drapisarda: I can not put them together. The _isActive_ value is set by the _focusHandler_ method. In the first test, I'm mocking that method, so the _isActive_ value is not affected. So, I can make just one test from this two, but I would, however, have to trigger _focus_ twice. So, it's pointless to have just one test.

tests/unit/components/form/HdRange.spec.js
drapisarda: Same also here.

src/components/form/HdRange.vue
drapisarda: I do agree.

tests/unit/components/form/HdRange.spec.js
viniciuskneves: How about not mocking it on the first test?
You can just spy the `focusHandler` so you can test it is being called and the `isActive` state is set as well.
What do you think?

drapisarda: I changed the order and it works good!

package.json
```diff
@@ -87,5 +87,13 @@
       "vue-cli-service lint",
       "git add"
     ]
+  },
+  "prettier": {
```
viniciuskneves: I don't think we've prettier setup in this project. Do you intend to do so? I would definitely recommend to do in another PR as it might update a lot of files all at once.

src/components/form/HdRange.vue
```diff
@@ -78,7 +78,12 @@ export default {
       };
     },
     stepsAmount() {
-      return 1 + (this.maxValue - this.minValue) / this.rangeStep;
+      const rangeSize = this.maxValue - this.minValue;
```
viniciuskneves: You can make `rangeSize` a computed property by itself.

src/stories/HdRange.stories.js
```diff
@@ -58,4 +58,25 @@ storiesOf('Form/HdRange', module)
...
     },
+  }))
+  .add('without steps', () => ({
+    components: { HdRange },
+    template: `
+      <div>
+        <HdRange
+          :minValue="0"
+          :maxValue="100"
+          :rangeStep="10"
```
viniciuskneves: It has steps and the story says it doesn't 🤔

drapisarda: FUUUUU! Bad description. It was actually without bullets. I made both the story and make the display bullet optional to make it behave more like the native one.

src/components/form/HdRange.vue
drapisarda: 👍

package.json
drapisarda: 👍

viniciuskneves:
Hey @drapisarda there are some conflicts in the `package-lock` file. You can run `npm install --package-lock-only` to solve it.
What I actually wanted to say is: don't force-push when there are reviews pending as it messes with the commit history and the comments itself. Just as a side note and suggestion of course.
Going to go through it once again today.

PR #178, opened 2019-10-17
[F] Add HdAlert

ilyasmez:
A component to show useful messages to the user.

@OriginalEXE, you can check how we are using the component in the login page [Login.vue](https://github.com/homeday-de/realtor-app/blob/develop/src/views/Login.vue) and maybe use it in myHomeday as well :)

I've already discussed with Andrea, and once the PR is merged I'll share the stories with the design team and see if we can use this component with type `info` instead of `HdInfoBox`.

I think it's the consumer's responsibility to sanitize the html before passing it, so I did it on the story level and not the component's (up for discussion).

And yeah, I added mantis (#63B65B) to our color palette, because we are using it in multiple places (in myHomeday we even added it as a [global variable](https://github.com/homeday-de/customer-app/blob/develop/src/styles/variables.scss#L6), so if this PR gets merged we can remove it from there), and we are keeping this color in the new design system :)


src/stories/HdAlert.stories.js
```diff
@@ -0,0 +1,86 @@
...
+        message="An error message"
+      />
+    `,
+  }))
+  .add('Playground 🎛📝', () => ({
+    components: { HdAlert },
+    props: {
+      type: {
+        type: String,
+        default: select('type', ['', 'success', 'info', 'warning', 'error']),
```
viniciuskneves: To avoid it (if we add more options), I would create the types in a different file so they can be easily imported among other components/files.

tests/unit/components/HdAlert.spec.js
```diff
@@ -0,0 +1,60 @@
...
+    });
+
+    expect(wrapper.classes()).toContain(`alert--${TEST_TYPE}`);
+  });
+
+  test('renders a custom icon', () => {
+    const ICON_URL = 'https://placeholder.pics/svg/24x24';
+    wrapper.setProps({
+      icon: ICON_URL,
+      noIcon: false,
```
viniciuskneves: You don't need to explicitly set it here.

src/stories/HdAlert.stories.js
ilyasmez: 1- That makes sense, thanks for the suggestion! Please check the new commit. I also went with dynamic stories, so we don't have to add/remove a stories if we change the types.
2- That's true 😅 I removed that option.

tests/unit/components/HdAlert.spec.js
ilyasmez: I left it to make sure that the icon is visible, just in case the default props get changed in the `mount` call.

But I agree it's better to avoid this kind of dependencies in tests. (I removed it)

package.json
```diff
@@ -41,6 +41,7 @@
     "babel-jest": "^23.6.0",
     "core-js": "^2.6.10",
     "coveralls": "^3.0.7",
+    "dompurify": "^2.0.6",
```
SinisaG: I am not sure we need this one, if we don't sanitize. Probably we don't need to do it in story as well.

ilyasmez:
@OriginalEXE, slot is the first thing I had in mind and it will spare us the html sanitization issue (the consumer has to add `v-html` on their template or have their own components there), I just wanted to stay consistent with the other components, as we mostly use props instead of slots. But I'm definitely in for the slots idea.

@homeday-de/frontend-engineering, 2 thumbs up to this comment and I'll change it to slot, probably revisit the other components as well.
Sometimes it's hard to draw a line between using a prop or slot, but for now we can keep it simple: If the component has one main content input and the component doesn't have to know about its value, then we use **slot**. What do you think?

ilyasmez:
Slots it is then! I'll make the necessary modifications and request your review again :)

package.json
SinisaG: It's more about dangerous data (script tags, any malicious code, etc.). No, I agree, it's a good practice to treat any source as not trusted. :+1:

ilyasmez:
@OriginalEXE 
Indeed! :)
As mentioned above, you can find an example in the [Realtor-app's Login.vue](https://github.com/homeday-de/realtor-app/blob/develop/src/views/Login.vue), it flashes as well 😄

PR #188, opened 2019-10-22
[T] Add license and enhance metadata in package.json

SinisaG:
https://homeday.atlassian.net/browse/FET-93
https://homeday.atlassian.net/browse/FET-94

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,24 @@
 {
   "name": "homeday-blocks",
   "version": "2.5.0",
+  "description": "A Vue component library built by Homeday's frontend team.",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/homeday-de/homeday-blocks.git"
+  },
+  "keywords": [
+    "vue",
+    "javascript",
+    "component library",
+    "storybook",
+    "homeday-blocks",
+    "services"
+  ],
+  "bugs": {
+    "url": "https://github.com/homeday-de/homeday-blocks/issues"
+  },
+  "homepage": "https://blocks.homeday.dev",
+  "license": "MIT",
   "scripts": {
     "lint": "vue-cli-service lint",
     "build:lib": "vue-cli-service build --target lib --name homeday-blocks ./src/components/index.js",
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,13 @@
+# This is the list of Homeday-blocks authors
+#
+# This does not necessarily list everyone who has contributed code, to see the full list
+# of contributors, see the revision history in source control.
+# List is order alphabetically based on last name
+
+Ilyas Amezouar <ilyas.amezouar@homeday.de>
+Juan Denis <juan.denis@homeday.de>
+Sinisa Grubor <sinisa.grubor@homeday.de>
+Oleg Ivaniv
+Vinicius Kiatkoski Neves <vinicius.neves@homeday.de>
+Daniele Rapisarda <daniele.rapisarda@homeday.de>
+Ante Scepic <ante.sepic@homeday.de>
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Homeday
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -1,6 +1,19 @@
 {
   "name": "homeday-blocks",
   "version": "2.5.0",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/homeday-de/homeday-blocks.git"
+  },
+  "keywords": [
+    "vue.js",
```
SinisaG: If you have a better suggestion here, please shoot. :)

package.json
```diff
@@ -1,6 +1,26 @@
...
+    "Javascript",
+    "Component library",
+    "storbybook",
+    "homeday-block",
+    "services"
+  ],
+  "bugs": {
+    "url": "https://github.com/homeday-de/homeday-blocks/issues"
+  },
+  "homepage": "https://github.com/homeday-de/homeday-blocks#readme",
```
OriginalEXE: What about using `https://blocks.homeday.dev/` for the homepage? We have a quite nice welcome page now and it also links to the GitHub page. Since we will have a dedicated github link on npm anyway, might be more valuable to use this one for something else.

SinisaG: Was thinking about that as well. Agree, will update. :+1:

viniciuskneves: I'm with @OriginalEXE here as well.

SinisaG: Done. :+1:

package.json
```diff
@@ -1,6 +1,26 @@
 {
   "name": "homeday-blocks",
   "version": "2.5.0",
+  "author": "Homeday - Frontend Team",
```
viniciuskneves: Maybe make this guy public: https://github.com/orgs/homeday-de/teams/frontend-engineering
Or we can use https://docs.npmjs.com/files/package.json#people-fields-author-contributors and add the frontend team to the list.

package.json
```diff
@@ -1,6 +1,26 @@
...
+  "description": "A Vue component library built by Homeday's frontend team.",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/homeday-de/homeday-blocks.git"
```
viniciuskneves: Don't need `git+` here. Can be just `github:homeday-de/homeday-blocks` or the full GitHub URL.

package.json
viniciuskneves: just `vue` as the Vue repo also uses it

image

package.json
```diff
@@ -1,6 +1,26 @@
...
+  },
+  "keywords": [
```
viniciuskneves: I think it ignores capital letters

package.json
viniciuskneves: Or
image

package.json
SinisaG: Thanks, I'll update to full github url.

https://docs.npmjs.com/files/package.json#repository

(I'm adding links for my own reference ;) )

package.json
SinisaG: Good reference. :+1:

package.json
SinisaG: Thanks for pointing it out. I've removed the author from `package.json` and added AUTHORS file. 

There is also a helper for git, what's missing, which is also helpful (did not know about it)
https://github.com/homeday-de/homeday-blocks/community

SinisaG:
Thanks for feedback @OriginalEXE , @viniciuskneves . I've updated accordingly. If you'd like to have another look, please go ahead.

package.json
```diff
@@ -1,6 +1,24 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/homeday-de/homeday-blocks.git"
+  },
+  "keywords": [
+    "vue",
+    "javascript",
+    "component library",
+    "storbybook",
+    "homeday-block",
```
viniciuskneves: block or blocks here?

package.json
```diff
@@ -1,6 +1,24 @@
...
+  "description": "A Vue component library built by Homeday's frontend team.",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/homeday-de/homeday-blocks.git"
+  },
+  "keywords": [
+    "vue",
+    "javascript",
+    "component library",
+    "storbybook",
```
viniciuskneves: storybook?

SinisaG: Oh my, is like I was :beers: . Thanks for extra pair of :eye: .

OriginalEXE:
:+1: Looking good

SinisaG:
I'll merge this metadata beauty. ;)

PR #187, opened 2019-10-22
[Release] Remove HdInfoBox, Add HdAlert, HdButton, HdLink, extend HdInput behavior, update project metadata, update dependencies

SinisaG:
#### Breaking change
- Remove HdInfoBox #178 

#### Features 
- Add HdAlert #178
- Add HdButton #158 
- Add HdLink #176 
- Emit 'blur' and 'focus' from HdInput #200
- Watch type on HdInput #202
 
#### Tasks
- Add license and enhance metadata in package.json #188 
- Update demo branch #186 
- Upgrade dependencies #170 

Due to #186 we need to update cloudfront configuration once this is released. So I'm opening a PR already - to leave this note. ;)

PR #201, opened 2019-10-30
[T] Bump version

SinisaG:
Bumping version for #187 . 

@viniciuskneves : Do you want to include #200 ? I'll release tomorrow.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "homeday-blocks",
-  "version": "3.0.0",
+  "version": "4.0.0",
   "description": "A Vue component library built by Homeday's frontend team.",
   "repository": {
     "type": "git",
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,4 +10,4 @@ Sinisa Grubor <sinisa.grubor@homeday.de>
 Oleg Ivaniv
 Vinicius Kiatkoski Neves <vinicius.neves@homeday.de>
 Daniele Rapisarda <daniele.rapisarda@homeday.de>
-Ante Scepic <ante.sepic@homeday.de>
+Ante Sepic <ante.sepic@homeday.de>


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "homeday-blocks",
-  "version": "2.5.0",
+  "version": "3.0.0",
```
SinisaG: Is a major version bump, as we have breaking changes (HdInfoBox).

viniciuskneves:
I've just merged #200 🚀

viniciuskneves:
@SinisaG just opened this one and would appreciate to have it in the same bump https://github.com/homeday-de/homeday-blocks/pull/202

ilyasmez:
@SinisaG 
We already have a [release](https://github.com/homeday-de/homeday-blocks/releases/tag/3.0.0) with the version `3.0.0`, but it was released from develop branch 😕 
We can either delete it and release the new one immediately, or we bump to `4.0.0`.

I vote for the first option.

viniciuskneves:
I vote for first opinion hehehe

SinisaG:
> @SinisaG
> We already have a [release](https://github.com/homeday-de/homeday-blocks/releases/tag/3.0.0) with the version `3.0.0`, but it was released from develop branch 
> We can either delete it and release the new one immediately, or we bump to `4.0.0`.
> 
> I vote for the first option, but we have to merge the change logs.

Oh, you're right (https://github.com/homeday-de/homeday-blocks/blob/master/package.json#L3) - that also explains the conflicts in `package.json`. The first option is not so easy though, as you can't overwrite the published version on npm. `npm publish -f` has been deprecated - https://blog.npmjs.org/post/77758351673/no-more-npm-publish-f

We could `unpublish`, wait for 24 hours and then `publish` again. 

Taking this in consideration, I would opt for bumping to `4.0.0`, resolving conflicts and target master. We're dealing with problem/no problem, where out version numbers are getting pretty high. Is a consequence of lot's of changes / stabilizing it. But I think that's fine, as long as we follow semversoning correctly. 

What do you think @homeday-de/frontend-engineering ?

ilyasmez:
@SinisaG 

> We could additionally still delete version 3.0 from git versions and re-release it targeting master - as this is the state of master currently (just to be consistent :) ). Then release version 4.0.0 on top.

I prefer this solution :)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "homeday-blocks",
-  "version": "3.0.0",
+  "version": "4.0.0",
```
SinisaG: Should be correct now. :+1:

viniciuskneves: Hey @SinisaG you can use [`npm version`](https://docs.npmjs.com/cli/version) in this case. It automatically bumps `package.json` and `package-lock.json` for example. You can check an example here: https://github.com/VueFloripa/vuefloripa.github.io/blob/development/package.json#L8

With this merge we got a mismatch between both files.

PR #208, opened 2019-10-31
[Release] Travis conf

SinisaG:
This is just addition to #187 .


.travis.yml
```diff
@@ -25,3 +25,4 @@ deploy:
 branches:
   only:
   - develop
+  - master
```
SinisaG: We need this to actually deploy on `master`. :)

PR #219, opened 2019-11-13
[B] Fix icons reference

OriginalEXE:
I tried using the component and whoops :D For now I'll simply add the icons to the same path in our project, but we need to fix this :)


OriginalEXE:
Will do first thing tomorrow :+1:

PR #221, opened 2019-11-14
[T] Bump patch version

OriginalEXE:
To be merged after #220

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "homeday-blocks",
-  "version": "4.0.0",
+  "version": "4.0.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "homeday-blocks",
-  "version": "4.0.0",
+  "version": "4.0.1",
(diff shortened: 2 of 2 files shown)


OriginalEXE:
Yeah, in this case, that would work as well. It might happen sometimes that not everything from the `develop` branch should be released. Do you see any drawbacks in doing it like this?

ilyasmez:
Both will have the same effect, no drawbacks. It will just respect the flow of not merging directly to `master` but going through `develop` first (except special cases).
Yes there might be situations where we don't want to release everything from develop, but it's not the case here :D that's why it's just a suggestion, I don't have any objections regarding the route you're taking.

SinisaG:
Hey, 
yeah, the drawback is that `master` and `develop` can get out of sync and then there are conflicts to deal with, which can be hard, if there are lots of changes. 

I suggest merging `develop` in `master` and keep the usual flow. For cases where `develop` is not ready to be merged (`hotfixes`), you can brach of `master` directly do the change and merge it to `master` and release. But make sure to cherry-pick your commit and open a PR to develop also. That will keep branches in sync. Ideally we will not have lots of `hotfixes`, so should not be to much of a concern. We need to increase test coverage. :wink:

Issue #226, opened 2019-11-25
Refactor <HdLink> exposed modifiers


viniciuskneves:
Currently the modifiers are hard coded inside the component and replicated inside the story itself.
We should follow the approach from HdButton where we expose those modifiers via Object and reuse them to create all available stories.

PR #228, opened 2019-11-25
[F] Percy snapshots

viniciuskneves:
It adds Percy snapshots in the build process. It takes screenshots from every story so I added a param to skip the ones with playground and also some with dynamic data. My idea here is to not refactor all components to add screenshot capabilities.
I've also set the width of a specific component that has a different behavior for mobile only.

closes https://github.com/homeday-de/homeday-blocks/issues/226

Upgraded some dependencies to close dependabot PRs.

### DIFF
--- a/.travis.variables.sh
+++ b/.travis.variables.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+export ENV=test
+
+if [ "$TRAVIS_BRANCH" = "develop" ]
+  then
+    echo "Running STAGING build"
+
+    export ENV=staging
+    export PERCY_TARGET_BRANCH=develop
+fi
+
+if [ "$TRAVIS_BRANCH" = "master" ]
+  then
+    echo "Running PRODUCTION build"
+
+    export ENV=production
+    export PERCY_TARGET_BRANCH=master
+fi
+
+echo "ENV: $ENV"
+echo "PERCY_TARGET_BRANCH: $PERCY_TARGET_BRANCH"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,23 @@
 language: node_js
 node_js:
-- 10.16.0
+- 10.17.0
+
 cache: npm
+
 install:
 - npm ci
+
+before_script:
+- npm run lint:ci
+
 script:
 - npm run test:unit:ci
 - npm run build:storybook
+- npm run test:percy:ci
+
 after_success:
 - npm run coverage
+
 deploy:
   provider: s3
   access_key_id: AKIAI3XOIWNY6PBLTBQQ
@@ -22,6 +31,7 @@ deploy:
   upload-dir: master
   on:
     branch: master
+
 branches:
   only:
   - develop
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ node_js:
 
 cache: npm
 
+before_install:
+- source .travis.variables.sh
+
 install:
 - npm ci
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_script:
 script:
 - npm run test:unit:ci
 - npm run build:storybook
-- npm run test:percy:ci
+- npm run test:percy
 
 after_success:
 - npm run coverage
--- a/README.md
+++ b/README.md
@@ -58,6 +58,12 @@ To watch for changes while writing tests for a single component:
 npm run test:unit:watch ComponentName
 ```
 
+### Screenshot tests on Percy
+Make sure you've `PERCY_TOKEN` exported first. It can be obtained from Percy dashboard if you've access to it.
+```bash
+npm run test:percy
+```
+
 #### Gotchas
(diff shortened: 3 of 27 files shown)


.travis.variables.sh
```diff
@@ -0,0 +1,22 @@
...
+if [ "$TRAVIS_BRANCH" = "master" ]
+  then
+    echo "Running PRODUCTION build"
+
+    export ENV=production
+    export PERCY_TARGET_BRANCH=master
+fi
+
+echo "ENV: $ENV"
+echo "PERCY_TARGET_BRANCH: $PERCY_TARGET_BRANCH"
```
viniciuskneves: This file is responsible for setting the `PERCY_TARGET_BRANCH` which is the branch that Percy will use as base for snapshots.

viniciuskneves: By default it is set do `develop` on Percy dashboard.

.travis.yml
```diff
@@ -1,14 +1,26 @@
 language: node_js
 node_js:
-- 10.16.0
+- 10.17.0
```
viniciuskneves: Latest v10 release (Shall we go to v12 by now?)

README.md
```diff
@@ -1,4 +1,6 @@
-[![Build Status](https://travis-ci.com/homeday-de/homeday-blocks.svg?token=rxKqgsBkMR5FqNpvxBcx&branch=develop)](https://travis-ci.com/homeday-de/homeday-blocks) [![Coverage Status](https://coveralls.io/repos/github/homeday-de/homeday-blocks/badge.svg?t=Kbz7Vb)](https://coveralls.io/github/homeday-de/homeday-blocks)
+[![Build Status](https://travis-ci.com/homeday-de/homeday-blocks.svg?token=rxKqgsBkMR5FqNpvxBcx&branch=develop)](https://travis-ci.com/homeday-de/homeday-blocks)
+[![Coverage Status](https://coveralls.io/repos/github/homeday-de/homeday-blocks/badge.svg?t=Kbz7Vb)](https://coveralls.io/github/homeday-de/homeday-blocks)
+[![This project is using Percy.io for visual regression testing.](https://percy.io/static/images/percy-badge.svg)](https://percy.io/Homeday/homeday-blocks)
```
viniciuskneves: image

package.json
```diff
@@ -21,13 +21,15 @@
...
     "lint": "vue-cli-service lint",
+    "lint:ci": "vue-cli-service lint --no-fix",
     "build:lib": "vue-cli-service build --target lib --name homeday-blocks ./src/components/index.js",
     "build:storybook": "vue-cli-service storybook:build -c config/storybook",
     "serve:storybook": "vue-cli-service storybook:serve -p 6006 -c config/storybook",
     "test:unit": "vue-cli-service test:unit",
     "test:unit:snapshot": "vue-cli-service test:unit -u",
     "test:unit:watch": "npm run test:unit -- --watch",
     "test:unit:ci": "npm run test:unit -- --ci --coverage",
+    "test:percy": "if [ $PERCY_TOKEN ]; then percy-storybook; else echo 'Skipping test: PERCY_TOKEN is not defined'; fi",
```
viniciuskneves: We need to check for `PERCY_TOKEN` otherwise it breaks as the variable can't be found.

viniciuskneves:
> Looks good. 👍 I get the idea, is a bit hard to see what's skipped and what not from PR. Would it make sense to have a story and then one render(or how many are needed?) dedicated to Percy and rest we skip for screenshots? So we know to always add a case for Percy when writing stories. For that, I guess would be beneficial to reverse the condition and instead of `skip`, we would say `include`. :)

We can see how it grows for next stories. The rule of thumb is: have we rendered it before (previous story) and this new story is only changing its style --> We should skip, otherwise it should be rendered. I've skipped few of them so I think it is open for discussion. I will try to keep track of the next PR and see if it makes sense to reverse the logic and make it more strict (I do see the benefit of making it strict as it is the same of "writing tests" for that component, if you don't take screenshot it is "not adding tests").

Let's see if someone else adds a comment today otherwise I will move forward and see how it evolves.

Thanks for the feedback and thoughts.

SinisaG:
> > Looks good. +1 I get the idea, is a bit hard to see what's skipped and what not from PR. Would it make sense to have a story and then one render(or how many are needed?) dedicated to Percy and rest we skip for screenshots? So we know to always add a case for Percy when writing stories. For that, I guess would be beneficial to reverse the condition and instead of `skip`, we would say `include`. :)
> 
> We can see how it grows for next stories. The rule of thumb is: have we rendered it before (previous story) and this new story is only changing its style --> We should skip, otherwise it should be rendered. I've skipped few of them so I think it is open for discussion. I will try to keep track of the next PR and see if it makes sense to reverse the logic and make it more strict (I do see the benefit of making it strict as it is the same of "writing tests" for that component, if you don't take screenshot it is "not adding tests").
> 
> Let's see if someone else adds a comment today otherwise I will move forward and see how it evolves.
> 
> Thanks for the feedback and thoughts.

Sounds good. :+1: Let's give it a try and see/learn.

PR #244, opened 2019-12-05
[F] Picture element and source sets on HDGalleryMedia and HDGalleryCarousel

drapisarda:
⚠️  There may be bugs or buggy things.

In those components, is now possible to use

- Images with a source
- Images, each one with its own source set
- Picture elements with multiple sources
- Picture elements with multiple sources, each one with its own source set

⚠️ The documentation is still missing, but you can start to review the functionalities.

### DIFF
--- a/src/components/gallery/HdGalleryCarousel.vue
+++ b/src/components/gallery/HdGalleryCarousel.vue
@@ -30,12 +30,16 @@
             :style="sizerStyles"
             class="gallery-carousel__item__sizer"
           />
-          <div
-            :style="{
-              'background-image': `url('${item.thumbnail}')`,
-            }"
-            class="gallery-carousel__item__background"
-          />
+
+          <picture v-if="usePictureElement" class="gallery-carousel__item__picture">
+              <source v-for="(source, media) in item.thumbnailPictureSources"
+                :key="media"
+                :media="`(${media})`" :srcset="source"
+              >
+            <img :src="item.thumbnail" :alt="item.caption" :srcset="item.thumbnailSrcSet">
+          </picture>
+          <img v-else class="gallery-carousel__item__picture" :src="item.thumbnail" :alt="item.caption" :srcset="item.thumbnailSrcSet">
+
         </div>
       </flickity>
       <div class="gallery-carousel__pager">
@@ -104,6 +108,10 @@ export default {
       return {
         paddingTop: `${100 / this.aspectRatio}%`,
       };
+    },
+    usePictureElement() {
+      const currentItem = this.items[this.currentIndex];
+      return currentItem.thumbnailPictureSources && currentItem.thumbnailPictureSources !== {};
     },
   },
   watch: {
@@ -259,6 +267,10 @@ export default {
     border-radius: 2px;
     overflow: hidden;
     transition: box-shadow $time-s ease-in-out;
+    background-color: $wild-sand;
+    display: flex;
+    align-items: center;
+    justify-content: center;
 
     @media (min-width: $break-tablet) {
       width: calc(100% / 5);
@@ -280,19 +292,21 @@ export default {
       margin-right: $inline-m;
     }
 
-    &__background {
+    img {
+      width: 100%;
+      height: 100%;
+      object-fit: cover;
+    }
+
+    &__picture {
       position: absolute;
       top: 0;
       right: 0;
       bottom: 0;
       left: 0;
-      background-color: $wild-sand;
-      background-position: center;
-      background-repeat: no-repeat;
-      background-size: contain;
-      @media (min-width: $break-tablet) {
-        background-size: cover;
-      }
+      display: flex;
+      align-items: center;
+      justify-content: center;
     }
   }
 }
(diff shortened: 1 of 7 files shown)


viniciuskneves:
@drapisarda how about waiting for https://github.com/homeday-de/homeday-blocks/pull/228?

So at least we would have some screenshot testing in place and could see the impacts out of it.

drapisarda:
Ok!

drapisarda:
Done

src/components/gallery/HdGalleryCarousel.vue
```diff
@@ -105,6 +109,10 @@ export default {
         paddingTop: `${100 / this.aspectRatio}%`,
       };
     },
+    usePictureElement() {
+      const currentItem = this.items[this.currentIndex];
+      return currentItem.thumbnailPictureSources && currentItem.thumbnailPictureSources !== {};
```
viniciuskneves: Are you sure about this check?

src/components/gallery/HdGalleryMedia.vue
```diff
@@ -55,6 +59,9 @@ export default {
         paddingTop: `${100 / this.aspectRatio}%`,
       };
     },
+    usePictureElement() {
+      return this.item.pictureSources && this.item.pictureSources !== {};
```
viniciuskneves: Same in here

src/components/gallery/HdGalleryCarousel.vue
```diff
@@ -30,12 +30,16 @@
...
-            class="gallery-carousel__item__background"
-          />
+
+          <picture v-if="usePictureElement" class="gallery-carousel__item__picture">
+              <source v-for="(source, media) in item.thumbnailPictureSources"
+                :key="media"
+                :media="`(${media})`" :srcset="source"
+              >
+            <img :src="item.thumbnail" :alt="item.caption" :srcset="item.thumbnailSrcSet">
+          </picture>
```
viniciuskneves: Can't we just keep this block? As far as I've seen IE doesn't support `<picture>` but it would fallback to the ``. Could you confirm it?

src/stories/mocks/GALLERY_ITEMS.js
```diff
@@ -1,51 +1,79 @@
...
+    thumbnailPictureSources: {
+      'max-width: 900px': 'https://dummyimage.com/768x432?text=4+thumb+M+to+TP+DPR1x 1x, https://dummyimage.com/1536x864?text=4+thumb+TP+DPI2x 2x, https://dummyimage.com/2304x1296?text=4+thumb+TP3x 3x',
+    },
+    thumbnailSrcSet: 'https://dummyimage.com/612x344?text=4+thumb+DPR2x 2x, https://dummyimage.com/918x516?text=4+thumb+DPR3x 3x',
   }, {
-    image: 'https://via.placeholder.com/768x432?text=5',
-    thumbnail: 'https://via.placeholder.com/153x86?text=5',
-    caption: 'Living room',
+    caption: 'Unevent image',
+    image: 'https://dummyimage.com/1536x2864?text=5+uneven+image',
```
viniciuskneves: To avoid it from happening again, can't we have those images on our side?

src/components/gallery/HdGalleryCarousel.vue
SinisaG: You can try it: 

```javascript
var a = {}; 
console.log(a === {}); // false
```

For checking if object has no keys, this is not a good check. You might want to use something like this instead 

```javascript
Object.entries(obj).length === 0
```

or

```javascript
Object.keys(obj).length === 0
```

and if you want to be super sure, you can additionally check the constructor - but above is probably already :+1:

src/components/gallery/HdGalleryCarousel.vue
SinisaG: It is correct. I've also tested in IE11, to validate. :) I think he's handling the backwards compatibility with that - based on the `if` condition. Probably can be consolidated though.

src/components/gallery/HdGalleryCarousel.vue
drapisarda: 👍

src/components/gallery/HdGalleryMedia.vue
drapisarda: 👍

src/components/gallery/HdGalleryCarousel.vue
drapisarda: Exactly

src/stories/mocks/GALLERY_ITEMS.js
drapisarda: Yes, but in another PR

src/components/gallery/HdGalleryCarousel.vue
viniciuskneves: I would give it a try first to see if we would need this fallback anyway (@aym3nb might help in here)

aym3nb:
Tested in customer app, everything works 👏

Issue #203, opened 2019-10-30
Extract HdInputPassword from HdInput


viniciuskneves:
Currently [`HdInput`](https://github.com/homeday-de/homeday-blocks/blob/develop/src/components/form/HdInput.vue) has some extra logic on top to handle `password` type. It could be extracted into another component to make it clear and don't overwhelm `HdInput` with logic.
We're currently developing also `HdInputFormatter` that should be possible to handle formatting inside `HdInput` but it is build on top of `HdInput` to not add extra logic layer into it.

ladas-larry:
Would you close this one? @viniciuskneves

viniciuskneves:
Hey @ladas-larry, this issue was tackled by @aym3nb already.

Check #250 for more details.

Did you find any further issues/improvements here that you would like to suggest/work on?

Thanks for asking!

ladas-larry:
Hi Vinicius, I see, thanks!

I'm looking for an opportunity to contribute to this repo.

I have further questions regarding https://github.com/homeday-de/homeday-blocks/issues/281, please have a look there.

Or if you have any task at your hands, that can be assign as a new issue, please open it and will follow up. :)

viniciuskneves:
Hey @ladas-larry thanks for your interest!

I will check #281 with @drapisarda to see if it still relevant and if he can provide more info there.

I will take some time tomorrow to move some "private" issues to GitHub, we have a bunch of tickets on our Jira board.

Get back to you tomorrow =]

ladas-larry:
Hi @viniciuskneves,
how has been your investigation of the tickets?
Any updates on that?

viniciuskneves:
Hey @ladas-larry, wanna give it a shot for this topic (https://github.com/homeday-de/homeday-blocks/pull/402#discussion_r406207406)?

We are discussing about having a mixin implementation but we're not sure yet how it would look like or so. You can give it a try and we can discuss about it in the PR. What do you think?

PR #250, opened 2019-12-13
[T] Extract HdInputPassword from HdInput 

aym3nb:
https://github.com/homeday-de/homeday-blocks/issues/203

### DIFF
--- a/src/components/form/HdInputPassword.vue
+++ b/src/components/form/HdInputPassword.vue
@@ -1,0 +1,236 @@
+<template>
+  <div
+    :class="fieldClasses"
+    class="field field--input"
+  >
+    <img
+      v-if="icon"
+      :src="icon"
+      role="presentation"
+      class="field__icon"
+    >
+    <input
+      :autocomplete="autocomplete"
+      :value="value"
+      :type="currentType"
+      :id="name"
+      :name="name"
+      :placeholder="isActive && placeholder !== undefined ? placeholder : ''"
+      :required="required"
+      :autofocus="autofocus"
+      :disabled="disabled"
+      class="field__input"
+      ref="input"
+      @input="handleInput"
+      @focus="handleFocus"
+      @blur="handleBlur"
+    />
+    <label
+      v-if="label"
+      :for="name"
+      class="field__label"
+    >
+      {{ label }}
+    </label>
+    <p
+      v-if="error"
+      class="field__error"
+    >
+      {{ error }}
+    </p>
+    <p
+      v-else-if="helper"
+      class="field__error field__error--helper"
+      v-html="helper"
+    />
+    <span v-if="showVisibilityToggle"
+      class="field__visibilityToggle"
+      :class="{'field__visibilityToggle--visible': currentType === 'text'}"
+      @click="togglePasswordVisibility"
+    />
+    <span class="field__border"/>
+  </div>
+</template>
+
+<script>
+import merge from 'lodash/merge';
+import { getMessages } from 'hd-blocks/lang';
+
+export default {
+  name: 'HdInputPassword',
+  props: {
+    label: {
+      type: String,
+      default: '',
+    },
+    name: {
+      type: String,
+      required: true,
+    },
+    value: {
+      type: [String, Number],
+      default: '',
+    },
+    placeholder: {
+      type: String,
+      default: '',
+    },
+    required: {
+      type: Boolean,
+      default: false,
+    },
+    autocomplete: {
+      type: String,
+      default: 'on',
+    },
+    autofocus: {
+      type: Boolean,
+      default: false,
+    },
+    lang: {
+      type: String,
+      default: 'de',
+    },
+    texts: {
+      type: Object,
+      default: () => ({}),
+    },
+    disabled: {
+      type: Boolean,
+      default: false,
+    },
+    icon: {
+      type: String,
+      default: '',
+    },
+  },
+  data() {
+    return {
+      currentType: 'password',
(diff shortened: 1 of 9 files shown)


src/components/form/HdInputPassword.vue
```diff
@@ -0,0 +1,236 @@
...
+    },
+    placeholder: {
+      type: String,
+      default: '',
+    },
+    required: {
+      type: Boolean,
+      default: false,
+    },
+    autocomplete: {
```
SinisaG: Not sure we need all this options?

src/components/form/HdInputPassword.vue
```diff
@@ -0,0 +1,236 @@
...
+    :class="fieldClasses"
+    class="field field--input"
+  >
+    <img
+      v-if="icon"
+      :src="icon"
+      role="presentation"
+      class="field__icon"
+    >
+    <input
```
SinisaG: Can we reuse our `HdInput` here?

src/components/form/HdInputPassword.vue
aym3nb: yes, we don't need most of them, I just left them just in case.

src/components/form/HdInputPassword.vue
aym3nb: I didn't reuse HdInput because from what I understood from the issue is that we want to keep the password input logic out of it.

viniciuskneves: We can reuse `HdInput` and avoid all the duplicated code from methods/computed. The idea is to remove the logic from the `HdInput` with this visibility toggle thing and so on.

SinisaG: :+1: 

Can you update accordingly @aym3nb ? (when you can :))

tests/unit/components/form/HdInputPassword.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    expect(wrapper.find('input').attributes().type).toBe('password');
+    expect(toggle.classes()).not.toContain(PASSWORD_VISIBLE_CLASS);
+  });
+
+  test('doesn\'t show visibility toggle if empty', () => {
+    wrapper.setProps({ value: '' });
+    expect(wrapper.find(VISIBILITY_TOGGLE_SELECTOR).exists()).toBe(false);
+  });
+
+  test('Supports disabling', () => {
```
viniciuskneves: It is also part of `<HdInput>`.

tests/unit/components/form/HdInputPassword.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    wrapper.setProps({
+      disabled: true,
+    });
+    expect(wrapper.find('input').attributes().disabled).toBe('disabled');
+  });
+
+  test('Can render an icon', () => {
```
viniciuskneves: It is also part of `<HdInput>`.

tests/unit/components/form/HdInputPassword.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    wrapper.setProps({ value: TEST_VALUE });
+    expect(wrapper.find('input').element.value).toBe(TEST_VALUE);
+  });
+
+  test('emits `input` event with the right payload', () => {
+    wrapper.find('input').setValue(TEST_VALUE);
+    expect(inputWrapper.emitted('input')[0][0]).toEqual(TEST_VALUE);
+  });
+
+  test('validates requiredness', () => {
```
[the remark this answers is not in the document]
aym3nb: got it thanks for the review!

src/components/form/HdInputPassword.vue
```diff
@@ -0,0 +1,92 @@
...
+import HdInput from 'hd-blocks/components/form/HdInput.vue';
+
+export default {
+  name: 'HdInputPassword',
+  components: {
+    HdInput,
+  },
+  props: {
+    ...HdInput.props,
+  },
```
viniciuskneves: It is just something that came up to my mind: is it for helping with VSCode validation or so? 
Correct me if I'm wrong but looks like you want to do something like: https://vuejs.org/v2/guide/components-props.html#Disabling-Attribute-Inheritance

src/components/form/HdInputPassword.vue
```diff
@@ -0,0 +1,92 @@
...
+  computed: {
+    showVisibilityToggle() {
+      return this.currentValue.length !== 0;
+    },
+  },
+  watch: {
+    value() {
+      this.currentValue = this.value;
+    },
+  },
```
viniciuskneves: Correct me if I'm wrong again, please, but looks like you're doing it jus to bind to `<HdInput>` the `v-model` from `<HdInputPassword>`, right? Is it the case? Or there is something that I'm missing here?

src/components/form/HdInputPassword.vue
aym3nb: Actually I'm doing the opposite because here I want to inherit all props from HdInput which is a child in this case.

viniciuskneves: Yep but do you need it? I mean, you're going to forward it all to `<HdInput>`, right?

aym3nb:
Thanks for the review @ilyasmez 
I updated the code accordingly with a small difference to avoid the error (Vue Warn: Avoid Mutating a Prop Directly)

ilyasmez:
@aym3nb Niiiiice! Thank you for the fast changes! ❤️
Nice handling of the warning, I didn't notice that one 👍

PR #254, opened 2019-12-13
[F] Prettier settings

drapisarda:


viniciuskneves:
Where is the prettier setup? You need to add its dependency. Recommend using vue-cli-plugin for that.
Nevertheless I would also bring to our team sync since it might have a big impact in the project and might also be considered in the other projects as well.
What do you think?

viniciuskneves:
@drapisarda are you still up to testing it?

PR #255, opened 2019-12-16
[B] Ignore LastPass extension from auto filling inputs

aym3nb:
LastPass breaks UX, especially with our collapsible forms.
Ignoring it can be done easily by adding this attribute : **data-lpignore="true"**
image

### DIFF
--- a/src/components/form/HdInput.vue
+++ b/src/components/form/HdInput.vue
@@ -21,6 +21,7 @@
       :disabled="disabled"
       class="field__input"
       ref="input"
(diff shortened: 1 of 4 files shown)


viniciuskneves:
I don't use LastPass but won't it block users to autofill their passwords/data directly from LastPass?

aym3nb:
@viniciuskneves no, actually we are just preventing it from showing those icons on the inputs themselves. but the extension can still be used from the extensions bar without any issues.

PR #259, opened 2019-12-20
[F] HDLazyImage as picture with srcset

drapisarda:
The title is pretty self-descriptive.

### DIFF
--- a/tests/unit/components/HdLazyImage.spec.js
+++ b/tests/unit/components/HdLazyImage.spec.js
@@ -3,21 +3,28 @@ import HdLazyImage from '@/components/HdLazyImage.vue';
 import ITEMS from '@/stories/mocks/GALLERY_ITEMS';
 
 describe('HdLazyImage', () => {
-  it('renders with the image as background', () => {
-    const wrapper = mount(HdLazyImage, {
-      src: ITEMS[0].image,
-      background: true,
-    });
+  let wrapper;
 
+  beforeEach(() => {
+    wrapper = mount(HdLazyImage, {
+      propsData: {
+        src: ITEMS[0].image,
+        background: true,
+      },
+    });
+  });
+
+  it('renders the image as background of a div', () => {
     expect(wrapper.html()).toMatchSnapshot();
   });
 
   it('renders with the image as picture element, sources and source set', () => {
-    const wrapper = mount(HdLazyImage, {
+    wrapper.setProps({
       src: ITEMS[3].image,
       srcset: ITEMS[3].imageSrcSet,
       srcSmall: ITEMS[3].thumbnail,
       pictureSources: ITEMS[3].pictureSources,
+      background: false,
     });
 
     expect(wrapper.html()).toMatchSnapshot();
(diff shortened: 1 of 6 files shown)


tests/unit/components/__snapshots__/HdLazyImage.spec.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`HdLazyImage renders with the image as background 1`] = `<picture> <img class="lazy-image" src=""></picture>`;
```
viniciuskneves: Shouldn't this guy render a `<div>` then?

SinisaG: Great cache @viniciuskneves . It should, both snapshots are rendered empty (no data.)

tests/unit/components/HdLazyImage.spec.js
```diff
@@ -0,0 +1,25 @@
+import { mount } from '@vue/test-utils';
+import HdLazyImage from '@/components/HdLazyImage.vue';
+import ITEMS from '@/stories/mocks/GALLERY_ITEMS';
+
+describe('HdLazyImage', () => {
+  it('renders with the image as background', () => {
+    const wrapper = mount(HdLazyImage, {
```
SinisaG: When you're passing props, you have to pass them under `propsData` key. 

Like so: 

```javascript
    const wrapper = mount(HdLazyImage, {
      propsData:{
        src: ITEMS[0].image,
        background: true,
      }
    });
```

Otherwise you're snapshots will render empty (see under comment by @viniciuskneves )

tests/unit/components/HdLazyImage.spec.js
```diff
@@ -0,0 +1,25 @@
...
+      src: ITEMS[0].image,
+      background: true,
+    });
+
+    expect(wrapper.html()).toMatchSnapshot();
+  });
+
+  it('renders with the image as picture element, sources and source set', () => {
+    const wrapper = mount(HdLazyImage, {
+      src: ITEMS[3].image,
```
SinisaG: Same here, props have to be passed under `propsData`

drapisarda: Done!

drapisarda:
> This component is supposed to _lazy load_ images, but I don't see any lazy loading implemented for the responsive images. If you want to have responsive images, you can just use that simple component with `HdResponsive` like so:
> 
> ```
> <HdResponsive #default="{ matches }">
>   <HdLazyImage v-if="matches.m" src="mobile.jpg" src-small="mobile-small.jpg" />
>   <HdLazyImage v-else src="default.jpg" src-small="default-small.jpg" />
> </HdResponsive>
> ```
> 
> Am I overlooking something?

Actuali it is implementing the LazyLoading via the Image object on the line 78

ilyasmez:
> Actually it is implementing the LazyLoading via the Image object on line 78

But only the default image is lazy loaded (for example in the story, you have to go beyond 1200px to see it). What if the user opens the app on a smaller screen? They will have no lazy loading.
And `<picture>` with multiple `<source>`s gives us the possibility to even show a completely different image for each breakpoint, so we will need a small version of each image if we want to cover that case, and we have to detect when the image changes (due to the viewport change) so we can lazy load the appropriate image.

If we want to cover all the cases above I think it will be better to have it in a separate component and maybe go with the suggested solution above instead of `<picture>` to keep it simple.

drapisarda:
So I'd say all of this it's pretty useless 😞

PR #274, opened 2020-01-06
[B] current item index doesn't update on touch slide

aym3nb:
The current item index wasn't being updated in HdGallery if the user made the slide with a touch (mobile) because no input event was emitted.


aym3nb:
@viniciuskneves Added, thanks for the review 👍

PR #275, opened 2020-01-08
[F] Gallery Tiles can use images instead of backgrounds

drapisarda:
With a parameter, we can have images elements instead of backgrounds for HdGalleryTiles

### DIFF
--- a/tests/unit/components/HdLazyImage.spec.js
+++ b/tests/unit/components/HdLazyImage.spec.js
@@ -1,0 +1,27 @@
+import { mount } from '@vue/test-utils';
+import HdLazyImage from '@/components/HdLazyImage.vue';
+import ITEMS from '@/stories/mocks/GALLERY_ITEMS';
+
+describe('HdLazyImage', () => {
+  let wrapper;
+  const sampleImage = ITEMS[2];
+
+  beforeEach(() => {
+    wrapper = mount(HdLazyImage, {
+      propsData: {
+        src: sampleImage.image,
+        srcSmall: sampleImage.thumbnail,
+        srcset: sampleImage.imageSrcSet,
+        background: false,
+      },
+    });
+  });
+
+  it('renders component', () => {
+    expect(wrapper.html()).toMatchSnapshot();
+  });
+
+  it('It fist display the thumbnail/small image', () => {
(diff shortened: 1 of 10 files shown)


SinisaG:
Agree, please add tests for `srcSet`. It will be very helpful, as we use this a lot.

drapisarda:
In HDLazyImage, I'm trying to test the Image load. My issue is the t the loading is detected using an Image object (:62) that I can not access via jest. Any idea on how to do this? 

Probably the Image object should accessible from the outer scope.

SinisaG:
@drapisarda : One way that comes to mind is to expose the reference to the image. And then call `onload` manually. 

e.g.: 

```javascript
  ....
  data() {
    return {
      imageLoaded: false,
      img: null,
    };
  },
 ....
    maybeLazyLoad() {
      this.setImageUrl('');

      if (!this.srcSmall) {
        this.setImageUrl(this.src);
        this.imageLoaded = true;
        return;
      }

      this.imageLoaded = false;
      this.setImageUrl(this.srcSmall);

      this.img = new Image();

      this.img.onload = () => {
        this.imageLoaded = true;
        this.setImageUrl(this.src, this.srcset);
      };

      this.img.src = this.src;
    },
```

And your test could look something like that: 
```javascript
  it('Test onload callback', async () => {
    wrapper.vm.img.onload()
    console.log(wrapper.html());
  });
```

I'll think about possible alternatives way. But this is pretty simple and not overwriting anything. I guess, as you'd like to tests the outcome of `onload` that could be fine.

drapisarda:
I removed all the srcset part to keep this PR scoped.

tests/unit/components/HdLazyImage.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    });
+  });
+
+  it('renders component', () => {
+    expect(wrapper.html()).toMatchSnapshot();
+  });
+
+  it('It fist display the thumbnail/small image', () => {
+    expect(wrapper.attributes().src).toBe(sampleImage.thumbnail);
+  });
```
viniciuskneves: Aren't those tests similar? I think you can condense both in just one test: snapshot + src.

drapisarda: Yes. I could actually just keep the toMatchSnapshot one.

PR #276, opened 2020-01-10
[F] HdBadge

aym3nb:
Useful badge component


src/components/HdBadge.vue
```diff
@@ -0,0 +1,217 @@
...
+      font-size: 18px;
+      line-height: 28px;
+      border-radius: 4px;
+      padding: $stack-xs $inline-s;
+      cursor: default;
+
+      &::after{
+        content: ' ';
+        display: none;
+        background: url('~hd-blocks/assets/icons/ic_chevron-down--small.svg');
```
ilyasmez: We can use the chevron we already have here (it won't make a big difference UI-wise):
```
background: url('~hd-blocks/assets/icons/ic_chevron-down.svg') no-repeat;
width: 16px;
height: 16px;
```

src/components/HdBadge.vue
```diff
@@ -0,0 +1,217 @@
...
+    &--collapsed{
+      display: block;
+    }
+
+    &--collapsed > &__label{
+      border-bottom-left-radius: 0;
+      border-bottom-right-radius: 0;
+
+      &::after{
+        background: url('~hd-blocks/assets/icons/ic_chevron-up.svg');
```
ilyasmez: And here we can just rotate it
```suggestion
        transform: rotateZ(180deg);
```

PR #289, opened 2020-01-21
[T] import mixins.scss explicitly in each component

aym3nb:
Some components don't explicitly reference mixins. This is not visible, as we deal with it in our projects like that: https://github.com/homeday-de/customer-app/blob/develop/vue.config.js#L56-L61 

However, this is undocumented magic.   And this should not be necessary. 

Make components more robust, by importing mixins explicitly in each component. They don't cost anything, as they don't output CSS.

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -13,14 +13,15 @@ module.exports = {
         fix: true,
       });
   },
-  css: {
-    loaderOptions: {
-      sass: {
-        prependData: `
-          @import "hd-blocks/styles/mixins.scss";
-        `,
-      },
-    },
-  },
   transpileDependencies: ['vue-zoomer'],
+  // uncomment to import mixins.scss implicitly in all components
+  // css: {
+  //   loaderOptions: {
+  //     sass: {
+  //       prependData: `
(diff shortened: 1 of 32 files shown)


vue.config.js
```diff
@@ -13,14 +13,15 @@ module.exports = {
...
+  // uncomment to import mixins.scss implicitly in all components
+  // css: {
+  //   loaderOptions: {
+  //     sass: {
+  //       prependData: `
+  //         @import "hd-blocks/styles/mixins.scss";
+  //       `,
+  //     },
+  //   },
+  // },
```
viniciuskneves: I would remove those commented lines as we want to get rid of it.

SinisaG:
> I'd personally prefer to go with this solution first (with alias), and when we move to the relative paths we do it at once in a separate PR. We have a ticket to "investigate" in that topic, so we are not 100% sure if relative paths will be the outcome of this investigation.
> And also to avoid inconsistencies in our code, we are using the alias everywhere currently.

Agree, let's deal with this, once we have decision about aliases. :+1:

PR #290, opened 2020-01-23
[F] Support custom validators for google autocomplete input

mlande:
- I noticed this use case while implementing validation for location searches. I had to programmatically set errors to the input reference. 

- The idea is to allow to declaratively configure validations for an input like so:
```
  props: {
        ...
        customValidators: [
          (value) => ({ isValid: value.length >= 10, errorMessage: 'should be at least ten characters long' })
       ],
      },
```

- Please let me know your thoughts. If you think it's a good idea, I'm happy to add async validation support as well as support the generic  `HDInput`

### DIFF
--- a/src/components/form/HdGoogleAutocomplete.vue
+++ b/src/components/form/HdGoogleAutocomplete.vue
@@ -92,6 +92,9 @@ export default {
     customRules: {
       type: Array,
       default: () => [],
+      validator: rulesProvided => rulesProvided.every(
+        ({ validate, errorMessage }) => typeof validate === 'function' && typeof errorMessage === 'string',
+      ),
     },
   },
   data() {
--- a/src/components/form/HdGoogleAutocomplete.vue
+++ b/src/components/form/HdGoogleAutocomplete.vue
@@ -255,16 +255,20 @@ export default {
     validate() {
       if (this.required && this.isEmpty) {
         this.showError(this.t.FORM.VALIDATION.REQUIRED);
+      } else if (this.customRules.length && !this.isEmpty) {
+        this.validateCustomRules();
       } else {
-        const firstFailingRule = this.customRules.find(({ validate }) => !validate(this.value));
-
-        if (firstFailingRule) {
-          this.showError(firstFailingRule.errorMessage);
-        } else {
-          this.hideError();
-        }
+        this.hideError();
       }
       return !this.error;
+    },
+    validateCustomRules() {
+      const firstFailingRule = this.customRules.find(({ validate }) => !validate(this.value));
+      if (firstFailingRule) {
+        this.showError(firstFailingRule.errorMessage);
+      } else {
+        this.hideError();
+      }
     },
   },
 };
--- a/src/components/form/HdGoogleAutocomplete.vue
+++ b/src/components/form/HdGoogleAutocomplete.vue
@@ -89,7 +89,7 @@ export default {
       type: Boolean,
       default: false,
     },
-    customValidators: {
+    customRules: {
       type: Array,
       default: () => [],
     },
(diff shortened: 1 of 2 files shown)


tests/unit/components/form/HdGoogleAutocomplete.spec.js
```diff
@@ -67,4 +67,61 @@ describe('HdGoogleAutocomplete', () => {
...
+      wrapper.vm.validate();
+
+      expect(wrapper.vm.isValid).toBe(false);
+      expect(wrapper.vm.error).toBe(atLeastTenCharsErrorMessage);
+
+      // valid
+      wrapper.setProps({
+        customValidators: [atLeastTenChars],
+        value: 'this one is long enough',
+      });
```
viniciuskneves: It is not necessary to set `customValidators` again here, right?

tests/unit/components/form/HdGoogleAutocomplete.spec.js
```diff
@@ -67,4 +67,61 @@ describe('HdGoogleAutocomplete', () => {
...
+      wrapper.vm.validate();
+
+      expect(wrapper.vm.isValid).toBe(false);
+      expect(wrapper.vm.error).toBe(startsWithSpaceErrorMessage);
+
+      // first one valid, second one invalid
+      wrapper.setProps({
+        customValidators: [startsWithSpace, atLeastTenChars],
+        value: ' short',
+      });
```
viniciuskneves: Same thing here.

src/components/form/HdGoogleAutocomplete.vue
```diff
@@ -249,7 +253,15 @@ export default {
       if (this.required && this.isEmpty) {
         this.showError(this.t.FORM.VALIDATION.REQUIRED);
       } else {
```
ilyasmez: What about the non-required and empty fields, probably we don't want to run the customValidators on them, do you think we should add this condition to each customValidator or just handle that case in a generic way:

```js
if (this.required && this.isEmpty) {
  // Show is required error
} else if (!this.isEmpty) {
  // Check the custom validators
} else {
  // just hide the errors
}
```

tests/unit/components/form/HdGoogleAutocomplete.spec.js
mlande: @viniciuskneves Technically not, you are right :)

I left it because I thought it makes the test setup more explicit and clear. "It validates these rules against this value". I'm ok with removing it as well

src/components/form/HdGoogleAutocomplete.vue
```diff
@@ -249,7 +253,15 @@ export default {
...
-        this.hideError();
+        const [firstFailingValidation] = this.customValidators
```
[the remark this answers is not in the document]
mlande: Good point, I will adapt it. 

If we ever want to support multiple validation error messages for an input field, we can just change it back later.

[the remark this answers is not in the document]
mlande: This will not work as suggested, though, since `firstFailingValidation` will end up being the _function_ and not the result.

This made me realize a problem in the abstraction – to get the error message, you need to invoke the function. Instead I suggest to change it to `customRules` which look like this:

```
[
  { 
    validate: value => value.length >= 10, 
    errorMessage: atLeastTenCharsErrorMessage 
  },
  // ...more rules
]
```
and then:
```
const firstFailingRule = this.customRules.find(({ validate }) => !validate(this.value));
...
...
this.showError(firstFailingRule.errorMessage);
```

See my latest changes for this new implementation. @ilyasmez thoughts? :)

src/components/form/HdGoogleAutocomplete.vue
```diff
@@ -89,6 +89,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    customValidators: {
+      type: Array,
+      default: () => [],
```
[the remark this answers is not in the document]
mlande: Absolutely! I've forgotten this Vue feature!

src/components/form/HdGoogleAutocomplete.vue
mlande: Good catch. I think handling it generically is the way to go. I will add some logic

mlande:
@viniciuskneves @ilyasmez I added some changes based on your comments and other thoughts I had. Feel free to take another look before we merge. 

I will create separate tickets for:

1. Support custom rules with async validation
2. Support similar functionality in `HDInput` component

src/components/form/HdGoogleAutocomplete.vue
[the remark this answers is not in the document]
ilyasmez: I had the same feeling about the function, nice that you fixed that as well! :D

mlande:
For documentation purposes, these are the further tickets created:

- [https://homeday.atlassian.net/browse/FET-119

](url)

- [https://homeday.atlassian.net/browse/FET-120](url)

PR #296, opened 2020-01-27
[T] Remove alias usage

viniciuskneves:
- Remove usage of webpack-alias inside components --> Left it for storybook --> Would be nice to remove it as soon as we upgrade to storybook 5.3 which introduces some changes
- Components can be imported as `import { HdLink } from 'homeday-blocks';` or `import HdLink from 'homeday-blocks/src/components/HdLink.vue';`

- [x] Update docs usage

Question:
I'm thinking about refactoring all the tests as stories so we can guarantee that we import it from `main.js` which means that the component is exported at least. Any inputs here?


src/styles/inputs.scss
```diff
@@ -41,14 +41,14 @@ $focusedLabelTop: 5px;
 
   &--empty {
     &:not(#{$f}--active):after {
-      background-image: url('~hd-blocks/assets/icons/ic-error.svg');
+      background-image: url('/assets/icons/ic-error.svg');
```
ilyasmez: This and all the the paths in this file won't work.
`/assets/icon.svg` with point to `PATH/assets/icon.svg` (`PATH == "blocks.homeday.de"` in the case of storybook for example), which is not what we want, we should use a relative path to the file locally.
The issue we will face here if we use `../assets/` like in the other case, is that this file is imported in other components, which means that the path will be resolved based on the location of the component where it's imported.
```
/assets
/directory1
  |_Component1.vue
  |_/directory2
     |_Component2.vue
```
So in the example above the icon will work on `Component1` but not on `Component2`.

viniciuskneves: The icon works for both components because I'm not using relative paths inside `inputs.scss` so once imported it won't be relative.

It is indeed a limitation from Sass in which it doesn't respect the relative paths from the imported component.

The `PATH` will be the root path used for the project in which for blocks it will be `blocks.homeday.de` but for every other project it will follow its root path.

Nevertheless it will work as expected as it will be built by the "father" project which will handle those files during build.

Example of a built project using this new structure:
image
image

To be honest I don't like this solution that much as it is kind of "magical". There is this guy: https://github.com/bholloway/resolve-url-loader but it will be necessary to add it to the "father" project anyway.

@homeday-de/frontend-engineering thoughts here?

ilyasmez: `ic_visibility-on.svg` for example will work because it's defined as `../../assets/icons/ic_visibility-on.svg` inside HdInputPassword, but the example above (`ic-error.svg`) won't work, the `/assets/icons/ic-error.svg` won't be touched by webpack because it's not a path, and the browser will try to access it as a URL, like:
`YOUR_APP_BASE/assets/icons/ic-error.svg` which doesn't exist, instead we want something similar to: `YOUR_APP_BASE/img/ic-error.HASH.svg`

viniciuskneves: @ilyasmez as we've just aligned I'm going to use `homeday-blocks` as an alias so we can resolve it properly thanks to https://github.com/webpack-contrib/sass-loader#resolving-import-at-rules

Gonna rework on it soon!

viniciuskneves:
I've forced push to easily add what have been discussed!

main.js
```diff
@@ -0,0 +1,48 @@
...
+export { default as HdGallery } from './src/components/gallery/HdGallery.vue';
+export { default as HdGalleryCarousel } from './src/components/gallery/HdGalleryCarousel.vue';
+export { default as HdGalleryMedia } from './src/components/gallery/HdGalleryMedia.vue';
+export { default as HdGalleryOverlay } from './src/components/gallery/HdGalleryOverlay.vue';
+export { default as HdGalleryPlaceholder } from './src/components/gallery/HdGalleryPlaceholder.vue';
+export { default as HdGalleryTiles } from './src/components/gallery/HdGalleryTiles.vue';
+export { default as HdZoomerGallery } from './src/components/gallery/HdZoomerGallery.vue';
+
+// Notifications
+export { default as HdNotifications } from './src/components/notifications/HdNotifications.vue';
```
SinisaG: Should we export services here as well? Otherwise you'd import them as `homeday-blocks/src/services/SERVICE_NAME`. What do you think?

viniciuskneves: Could be! Gonna work on it!

viniciuskneves:
@homeday-de/frontend-engineering please ignore coverage drop in this case.

main.js
viniciuskneves: Could you check the exported services?

ilyasmez:
Looks good to me 👍 

I think exporting the services that way won't support the named export of the functions and will make us import the whole services instead of the functions we need. But that's a good first iteration.

We can think of importing the functions like so `import { fn } from ' homeday-blocks/service';` in a second iteration :)

main.js
```diff
@@ -0,0 +1,59 @@
...
+export { default as HdNotificationsBar } from './src/components/notifications/HdNotificationsBar.vue';
+
+// Services
+export { default as BreakpointsService } from './src/services/breakpoints';
```
SinisaG: :heart:

PR #297, opened 2020-01-28
[T] Document date service

SinisaG:
https://homeday.atlassian.net/browse/FET-84

This PR: 
- [x] Adds global config for jest (to fix the timezone)
- [x] Tests date service
- [x] Documents date service

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -5,6 +5,7 @@ module.exports = {
   globalSetup: './tests/global-setup.js',
   collectCoverageFrom: [
     'src/components/**/*.vue',
+    'src/services/**/*.js',
   ],
   coverageReporters: [
     'lcov',
--- a/jest.config.js
+++ b/jest.config.js
@@ -2,6 +2,7 @@
  * NOTE: hd-blocks mapper is needed to solve hd-blocks path in the components
  */
 module.exports = {
+  globalSetup: './tests/global-setup.js',
   collectCoverageFrom: [
     'src/components/**/*.vue',
   ],
--- a/src/services/README.md
+++ b/src/services/README.md
@@ -4,6 +4,7 @@ are here to be consumed in projects directly.
 
 Index:
 - [Breakpoints service](https://github.com/homeday-de/homeday-blocks/tree/develop/src/services#breakpoints) - Used for setting (and getting) the global project breakpoints.
+- [Date service](https://github.com/homeday-de/homeday-blocks/tree/develop/src/services#date) - Used as helpers to work with dates and date intervals.
 
 ## Breakpoints
 Exposes methods for setting up viewport breakpoints of the project. The breakpoints set in this way are later used by the `HdResponsive` component.
--- a/src/services/README.md
+++ b/src/services/README.md
@@ -68,3 +68,128 @@ It is worth noting that HdResponsive component takes a `breakpoints` prop as wel
 </HdResponsive>
 </template>
 ```
+
+## Date
+
+##### resetDateTime
+
+Returns date with hour, minutes, seconds and miliseconds set to `0`.
+`param {Date} date - arbitrary date`
+
+Example:
+
+```javascript
+const date = new Date(2018, 1, 15, 16, 30, 30, 30); // 15.2.2018 at 16:30:30:30
+const date2 = resetDateTime(d); // 15.2.2018 at 00:00:00:00
+```
+
+##### getNDaysFromDate
+
+Returns new date that i s N days away from inputted date, with hour, minutes, seconds and miliseconds set to `0`.
+
+`param {Date} initialDate - arbitrary date`
+`param {Number} amountOfDays - offet in days`
+
+```javascript
+const date = new Date(2018, 1, 15, 16, 30, 30, 30); // 15.2.2018 at 16:30:30:30
+const date2 = getNDaysFromDate(d, 1); // 16.2.2018 at 00:00:00:00
+```
+
+##### convertMsToDays
+
+Returns number of days from miliseconds.
+
+`param {Number} ms - miliseconds`
+
+```javascript
+const days = convertMsToDays(5 * 24 * 60 * 60 * 1000); // 5 days
+```
+
+##### getDaysDateRange
+
+Returns array of date range between 2 dates, including start date and *excluding* end date
+
+`param {Date} startDate - start of time interval`
+`param {Date} endDate - end of time interval`
+
+```javascript
+const timeInterval = getDaysDateRange(new Date(2018, 0, 1), new Date(2018, 0, 3));
+// [ 2018-01-01T00:00:00.000Z, 2018-01-02T00:00:00.000Z ]
+```
+
+##### getIntlDateString
+
+Returns array of *unique* day names in arbitrary(passed) date list
+
+`param {String} locale - Valid locale`
+`param {Array} datesArray - list of dates to be check for date`
+`param {Object} localStringArgs - Locale options - see options here for full list of options: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString#Parameters`
+
+
+```javascript
+const daysInIntervalLong = getIntlDateString('en-US', [new Date(2018, 0, 1)], { weekday: 'long' });
+// [ 'Monday' ]
+```
+
+##### generateRandomIndexes
+
+Returns array of N(count) indexes on `min` to `max` open interval
+
+`param {Number} amount - Total number of indexes to be generated`
+`param {Number} min - minimum value that indexes can have`
+`param {Number} max - minimum value that indexes can have`
+
+```javascript
+    const max = 5;
+    const min = 1;
+    const count = 3;
+    const randomInts = generateRandomIndexes(count, min, max);
+    // [2, 3, 5]
+```
+
+##### generateDateCycles
+
+Returns an array of dates in N (cycleLengthWeeks) weeks, with M cycles (amountOfCycles) from speficis offset from today (startDateOffset)
+
+`param {Number} cycleLengthWeeks - Total number of weeks in the cycle`
+`param {Number} amountOfCycles - Number of cycles to generate`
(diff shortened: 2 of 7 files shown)


jest.config.js
```diff
@@ -2,6 +2,7 @@
  * NOTE: hd-blocks mapper is needed to solve hd-blocks path in the components
  */
 module.exports = {
+  globalSetup: './tests/global-setup.js',
```
SinisaG: Injecting global setting to `jest`. This is executed before tests are run.

tests/global-setup.js
```diff
@@ -0,0 +1,3 @@
+module.exports = async () => {
+  process.env.TZ = 'UTC';
```
SinisaG: Fixing timezone, to avoid issues with testing.

tests/unit/services/date.spec.js
```diff
@@ -0,0 +1,138 @@
...
+    const daysInIntervalLong = getIntlDateString('en-US', datesList, { weekday: 'long' });
+    expect(daysInIntervalLong.length).toBe(7);
+    expect(daysInIntervalLong[0]).toBe('Sunday');
+
+    const daysInIntervalShort = getIntlDateString('en-US', datesList, { weekday: 'short' });
+    expect(daysInIntervalShort.length).toBe(7);
+    expect(daysInIntervalShort[0]).toBe('Sun');
+  });
+
+  it('generateRandomIndexes returns array of random integers on open interval', () => {
```
SinisaG: I am not sure why is this part of date service? Should I move it somewhere else? 

Also there's something strange in design here, that I am unsure about. If you look (here)[https://github.com/homeday-de/homeday-blocks/blob/develop/src/services/date.js#L45-L56] - to make it easier to discuss, I'm copying the relevant part of the code: 

```javascript
export const generateRandomIndexes = (amount, min, max) => {
  const indexes = [];

  while (indexes.length < amount) {
    const randomInt = getRandomInt(min, max);
    /* eslint-disable no-continue  */
    if (indexes.indexOf(randomInt) > -1) continue;
    indexes.push(getRandomInt(min, max));
  }

  return indexes;
};
```

What is the purpose of the `continue` statement? Initially I've read this as making sure we are returning unique indexes, however that does not seem to be the case as at the end, we don't push our original index, but generate new random value -> `indexes.push(getRandomInt(min, max));` . Anyone has more insights about implementation?  

This function also has potential to block the main threat if input params are not selected carefully. 
e.g.: 
```javascript
const amount = 100;
const min = 1;
const max = 2;
generateRandomIndexes (count, min, max);
```

That would loop....should we add some sanity checks at the beginning? I can add it in this PR. 

Anyway, if someone has more insights on actual usage here, I'll be happy to update this and document properly.

jest.config.js
```diff
@@ -2,8 +2,10 @@
...
   collectCoverageFrom: [
     'src/components/**/*.vue',
+    'src/services/**/*.js',
```
SinisaG: Adding services to coverage reporting.

SinisaG: That obviously dropped our overall coverage. :) I'm not going to cover all the services with tests in this PR, so I might loosen up rules for merging to get this in. :)

src/services/README.md
```diff
@@ -66,4 +67,152 @@ It is worth noting that HdResponsive component takes a `breakpoints` prop as wel
   <div v-else-if="matches.m">m breakpoint matched (global breakpoints still available since objects are merged)</div>
 </HdResponsive>
 </template>
+```
+
+## Date
```
SinisaG: You can see formatted docs here: https://github.com/SinisaG/homeday-blocks/tree/task/date-service/src/services#date
I guess it makes it a bit easier to read. :) Feedback here welcomed, about format, info, snippets,...

viniciuskneves:
image


https://coveralls.io/builds/28378777

tests/unit/services/date.spec.js
ilyasmez: If I'm not mistaken this is used to mock some charts data in one of our projects.
I agree with you `continue` is useless here, `randomInt` should be pushed to the array.

IMO this function seems too specific, I don't think we should have it in Blocks, or at least it should go to [utils.js](https://github.com/homeday-de/homeday-blocks/blob/develop/src/services/utils.js)

SinisaG: I think it fits `utils` more indeed. Probably renamed to `getNRandomNumbersOnTheInterval` or something shorter. :) 

I'll jobby it up and move it. Will also add sanity check to avoid issues.

tests/unit/services/date.spec.js
```diff
@@ -0,0 +1,138 @@
...
+    addedDayes = -1;
+    const yesterday = getNDaysFromDate(fullDate, addedDayes);
+    expect(yesterday.getFullYear()).toBe(year);
+    expect(yesterday.getMonth()).toBe(month);
+    expect(yesterday.getDate()).toBe(date - 1);
+    expect(yesterday.getHours()).toBe(0);
+    expect(yesterday.getMinutes()).toBe(0);
+    expect(yesterday.getSeconds()).toBe(0);
+    expect(yesterday.getMilliseconds()).toBe(0);
+  });
```
viniciuskneves: You could have a `describe` block in here and test those stuffs separately  as this test is quite extensive and the test title is big as well

tests/unit/services/date.spec.js
viniciuskneves: It is actually copied from this guy: https://github.com/homeday-de/sliders/blob/d1356d2e1200a794aaadf0ac73823513ca1210b3/src/services/dateService.js#L47

It is sliders specific and might have been copied as a whole. I'm completely fine about removing this guy from blocks and leave it for sliders.

tests/unit/services/date.spec.js
SinisaG: Done. :+1:  Does it look more readable?

src/services/utils.js
```diff
@@ -99,6 +99,27 @@ export function loadScript({ url = '', first = false, head = false }) {
   });
 }
 
+// Returns list of random integers
+export const generateRandomNumbers = (amount, min, max) => {
+  const indexes = [];
+  // sanity check to prevent infinit loop
+  if (amount > (max - min)) {
```
SinisaG: I've renamed this and refactored a bit.

SinisaG:
Thanks for feedback, I've refactored as per input. Please feel free to have another look @homeday-de/frontend-engineering . 

@jpDenisO : Can we start consuming `date` service from here in sliders instead of having a local copy? Please note the changes. :)

tests/unit/services/utils.spec.js
```diff
@@ -0,0 +1,27 @@
...
+  describe('generateRandomNumbers', () => {
+    it('returns array of random unique integers on open interval', () => {
+      const max = 5;
+      const min = 1;
+      const amount = 3;
+      const randomInts = generateUniqueNumbers(amount, min, max);
+      expect(randomInts.length).toBe(amount);
+      expect(Math.max(...randomInts)).toBeLessThanOrEqual(max);
+      expect(Math.min(...randomInts)).toBeGreaterThanOrEqual(min);
+      expect(new Set(randomInts).size).toBe(randomInts.length);
```
SinisaG: This is a check to validate numbers are indeed unique. Is it clear or a comment would be welcomed?

src/services/utils.js
```diff
@@ -99,6 +99,27 @@ export function loadScript({ url = '', first = false, head = false }) {
...
+  }
+  // max tries are arbitrary selected attempts limit
+  // to prevent super expensive cases of finding all
+  // unique numbers
+  let maxTries = 10 * amount;
+  while (indexes.length < amount && maxTries > 0) {
+    maxTries -= 1;
+    const randomInt = getRandomInt(min, max);
+    /* eslint-disable no-continue  */
+    if (indexes.indexOf(randomInt) > -1) continue;
```
ilyasmez: We can avoid disabling eslint here:
```js
if (!indexes.includes(randomInt)) {
  indexes.push(randomInt);
}
```

(I know it's from the old implementation 😅 )

src/services/README.md
```diff
@@ -66,4 +68,154 @@ It is worth noting that HdResponsive component takes a `breakpoints` prop as wel
...
+    2020-02-22T00:00:00.000Z,
+    2020-02-23T00:00:00.000Z,
+    2020-02-24T00:00:00.000Z
+  ]
+*/
+```
+
+## Utils
+
+#### generateUniqueNumbers
```
ilyasmez: Can we also mention that an empty or incomplete array is returned because of the sanity checks?

And also that both `min` and `max` are included, or maybe change "open interval" to "closed interval", and make it in bold or something because it's important information.

We should consider changing `getRandomInt` to not include `max`, to match the common reasoning.

tests/unit/services/utils.spec.js
SinisaG: :+1: 

Which `expects` do you mean? One is for `max`, one for `min` and one for `unique. `

src/services/utils.js
SinisaG: Yes! I realized after pushing, this should be refactored. :) Will updated. :+1:

src/services/README.md
SinisaG: Great points, will updated. :+1:

src/services/utils.js
SinisaG: Caught one more bug here, this should be `max-  min + 1`

SinisaG:
Thanks for the feedback, I've updated accordingly. Will leave it open for a bit more if any further feedback comes up and merge it in the afternoon. :+1:

SinisaG:
Going to merge this, even though coveralls don't pass - we know about this 1.3% drop, due to services being added. Don't think it makes sense to add bunch of tests in this PR. :)

viniciuskneves:
Sorry @SinisaG didn't have time go follow up here, just got time to read the comments once again here.

Nice job!

PR #298, opened 2020-01-28
[F] HdRange enhancements

ilyasmez:
### Changes:
- Added support of labels.
- Added support of tooltip (can be customized).
- Changed the naming of properties to match the native component.
- Added the possibility to change the track and progress bar backgrounds (of type CSS `<color>`).
- Update the UI on resize.
- Added stories and tests for the new features.
- Did some small optimizations, like: 
  - Saving the track width to avoid getting `offsetWidth` and updating it only when needed, instead of calling it each time a property change, which saves us some reflows.
  - Removed the bullets padding calculation, made the bullets clickable instead.
  - Change the properties default values to what makes more sense.

###### Because we needed the component in the valuation wizard, I started working on the component in a [separate branch](https://github.com/homeday-de/homeday-blocks/tree/feature/customizable-hd-range) before the original PR was merged, that's why you see some commits from @drapisarda in the history. That shouldn't have any drawbacks.


src/stories/HdRange.stories.js
```diff
@@ -80,22 +84,116 @@ storiesOf('Form/HdRange', module)
...
+    components: { HdRange },
+    template: `
+      <div>
+        <HdRange
+          v-model="currentValue"
+          :min="0"
+          :max="4"
+          :track-background="trackBackground"
+          :progress-background="progressBackground"
+          :labels="labels"
```
viniciuskneves:
```suggestion
```

You didn't define `labels` in the `props`

src/components/form/HdRange.vue
```diff
@@ -43,96 +77,125 @@ export default {
...
+      type: Boolean,
+      default: false,
+    },
+    tooltipValue: {
+      type: [String, Number],
+      default: '',
+    },
+    trackBackground: {
+      type: String,
+      default: 'fallback to the styles',
```
viniciuskneves:
```suggestion
      default: '',
```

No?

src/components/form/HdRange.vue
```diff
@@ -43,96 +77,125 @@ export default {
...
+    },
+    progressBackground: {
+      type: String,
+      default: 'fallback to the styles',
```
viniciuskneves:
```suggestion
      default: '',
```

Same here?

Issue #222, opened 2019-11-21
Rewrite HdResponsive to use MediaQueryList.addListener


OriginalEXE:
For the HdResponsive component, right now we rely on the resize event to check and update the matching breakpoints. It's throttled so there is no huge performance hit, but we could instead set up listeners on each media query, which fire only once (when the change happens).

More docs here: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener

It is well supported in all browsers we care about. The only "complexity" is that the recommendation is to use `addEventListener` instead of `addListener`, but `addEventListener` is not available in Safari. This can be solved by checking for existence and calling the appropriate one. Their functionality is the same.

drapisarda:
Working on this

drapisarda:
My proposal https://github.com/homeday-de/homeday-blocks/pull/299

PR #299, opened 2020-01-31
[F] MediaQueryList to manage HDResponsive

drapisarda:
This is my proposal to solve this
https://github.com/homeday-de/homeday-blocks/issues/222


src/components/HdResponsive.vue
```diff
@@ -18,42 +18,46 @@ export default {
...
+        try {
+          this.listeners[breakpointKey].addEventListener('change', (e) => {
+            this.updateMatch(breakpointKey, e.matches);
+          });
+        } catch (error) {
+          this.listeners[breakpointKey].addListener((e) => {
+            this.updateMatch(breakpointKey, e.matches);
+          });
+        }
+      }
```
viniciuskneves: To avoid this `for...in` I would go for something like `Object.entries(breakpoints).forEach(([key, value]) => {...})` so you can remove the ESLint ignore rules as well.

Doing `breakpoints[breakpointKey]` in the middle of the code is quite hard to read. So I would consider extracting it to variable or the solution I mentioned also touches it.

drapisarda: 👍  Als using that notation would remove the need of having `breakpoints[breakpointKey]` because we now have `value`

src/components/HdResponsive.vue
```diff
@@ -18,42 +16,46 @@ export default {
...
+  beforeMount() {
     this.setResponsiveFlags();
   },
-  mounted() {
-    this.addResizeEvents();
-  },
-  beforeDestroy() {
-    this.removeResizeEvents();
+  created() {
+    this.setResponsiveFlags();
```
viniciuskneves: Any special reason for calling it in `beforeMount` and `created`?

src/components/HdResponsive.vue
```diff
@@ -18,42 +16,46 @@ export default {
...
   },
-  created() {
+  beforeMount() {
     this.setResponsiveFlags();
   },
-  mounted() {
-    this.addResizeEvents();
-  },
-  beforeDestroy() {
-    this.removeResizeEvents();
```
viniciuskneves: I think we should consider: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/removeListener

We are not removing the listeners anymore, right?

src/components/HdResponsive.vue
drapisarda: I removed the _created_ one. It will come up with my next push

src/components/HdResponsive.vue
drapisarda: The listeners are data of this component so they should be demolished by the garbage collector after the destruction.

drapisarda:
I'm starting hating coverage

src/components/HdResponsive.vue
```diff
@@ -18,42 +16,46 @@ export default {
...
+  beforeMount() {
+    this.setListeners();
   },
   methods: {
-    setResponsiveFlags() {
+    /* istanbul ignore next */
+    updateMatch(key, value) {
+      this.matches[key] = value;
+      this.$forceUpdate();
```
viniciuskneves: Could you briefly describe the issue and why you went for `$forceUpdate`?

src/components/HdResponsive.vue
```diff
@@ -18,42 +16,46 @@ export default {
...
-    this.addResizeEvents();
-  },
-  beforeDestroy() {
-    this.removeResizeEvents();
+  beforeMount() {
+    this.setListeners();
   },
   methods: {
-    setResponsiveFlags() {
+    /* istanbul ignore next */
```
viniciuskneves: I don't think you need to ignore it in Istanbul in this case

src/components/HdResponsive.vue
```diff
@@ -18,42 +16,46 @@ export default {
...
-  created() {
-    this.setResponsiveFlags();
-  },
-  mounted() {
-    this.addResizeEvents();
-  },
-  beforeDestroy() {
-    this.removeResizeEvents();
+  beforeMount() {
+    this.setListeners();
```
viniciuskneves: Why did you go for `beforeMount` instead of `created` in this case?

src/components/HdResponsive.vue
drapisarda: Because it's not re-rendering at the matches change.

src/components/HdResponsive.vue
drapisarda: I need it to have _updateMatch_ not counted on the coverage. Unfortunately, I can no test it because of the differences between the CLI and the browser environment. Same for the listeners.

src/components/HdResponsive.vue
drapisarda: Does it make any big difference?

src/services/date.js
```diff
@@ -38,13 +38,13 @@ export const getIntlDateString = (locale, datesArray, localStringArgs) => datesA
   return accumulator;
 }, []);
 
-export const generateDateCycles = (cycleLengthWeeks, amountOfCycles, startDateOffset) => {
+export const generateDateCycles = (cycleLengthWeeks, amountOfCycles, startDateOffset, startDate = false) => {
```
viniciuskneves:
```suggestion
export const generateDateCycles = (cycleLengthWeeks, amountOfCycles, startDateOffset, startDate = new Date()) => {
```

src/services/date.js
```diff
@@ -38,13 +38,13 @@ export const getIntlDateString = (locale, datesArray, localStringArgs) => datesA
...
   const WEEK_DAYS = 7;
   const cycleLengthDays = cycleLengthWeeks * WEEK_DAYS;
-  const startDate = getNDaysFromDate(new Date(), startDateOffset);
-  const endDate = getNDaysFromDate(new Date(), (cycleLengthDays * amountOfCycles) + startDateOffset);
+  const calculatedStartDate = getNDaysFromDate(startDate || new Date(), startDateOffset);
```
viniciuskneves:
```suggestion
  const calculatedStartDate = getNDaysFromDate(startDate, startDateOffset);
```

src/services/date.js
```diff
@@ -38,13 +38,13 @@ export const getIntlDateString = (locale, datesArray, localStringArgs) => datesA
...
+  const endDate = getNDaysFromDate(startDate || new Date(), (cycleLengthDays * amountOfCycles) + startDateOffset);
```
viniciuskneves:
```suggestion
  const endDate = getNDaysFromDate(startDate, (cycleLengthDays * amountOfCycles) + startDateOffset);
```

src/components/HdResponsive.vue
viniciuskneves: Depends on your use case, but you decided to go from `created` to `beforeMount` so I'm wondering why...

drapisarda:
> As @viniciuskneves mentioned, please remove the changes that are not in the scope of this PR (`MediaQueryList to manage HDResponsive`).

This PR reduces the test coverage, So I have to cover something

viniciuskneves:
> > As @viniciuskneves mentioned, please remove the changes that are not in the scope of this PR (`MediaQueryList to manage HDResponsive`).
> 
> This PR reduces the test coverage, So I have to cover something

We've had other PRs that decrease test coverage as well and we approved it. If you add those tests in a different PR we will have test increase.
Nevertheless it shouldn't decrease coverage once you're updating `HdResponsive`. It should at least stay the same as you're not adding new components without tests or so.

I think you've rebased this PR as well, so try removing the new tests and see how the coverage behaves. Last time it was quite outdated...

src/components/HdResponsive.vue
viniciuskneves: @drapisarda ?

drapisarda: It just sounded better

viniciuskneves: If that is the case, I would keep it in `created` hook.

viniciuskneves:
Gonna give it a try in our app to see if I find any side effects but looks good.

PR #309, opened 2020-02-03
[T] Design system: Colors 

ilyasmez:
- I double-checked the abstraction with the design team.
- Took care of some hard coded values we had before, so the currently used colors should all be from the new DS.
- Removed the class and mixin generators, except increasing the size of our css they do nothing (up for discussion), I've already checked all of our projects and none of them uses the color classes.
- The only color that is still hardcoded is the black used for the shadows. I'll bring the topic of elevation to our upcoming DS sync.

#### Something to keep in mind:
The shades are based on the rounded value of the original color's lightness.
The reason behind this is if we take for example cello `#1c3553` it's lightness is ~21.8% which is not the same as the `22%` we are using for the shades.
`hsl(213, 50%, 22%)` converts to `#1c3554` 🤷‍♂️  (This was aligned with the design team)


ilyasmez:
@SinisaG Now that we have a test with random values, this [line](https://coveralls.io/builds/28501628/source?filename=src%2Fservices%2Futils.js#L116) might be missed sometimes. Shall we loosen up the coverage rules for the PRs? I'm getting an ❌ for a 0.06% decrease 😄

ilyasmez:
@homeday-de/frontend-engineering Just FYI, I changed DB to DB-110 as specified in the mapping sheet. I'm not a big fan of the idea tbh, because if we are using DB-110 instead of DB everywhere, then maybe we should use DB-110 as our secondary color, and not DB.
I've already let the design team know about my concerns. This is not a blocker for the PR though.

PR #316, opened 2020-02-10
[F] HdGallery emits on current item click

drapisarda:
A feature that I'm going to need soon.

### DIFF
--- a/tests/unit/components/HdGallery.spec.js
+++ b/tests/unit/components/HdGallery.spec.js
@@ -2,6 +2,8 @@ import { mount } from '@vue/test-utils';
 import HdGallery from '@/components/gallery/HdGallery.vue';
 import ITEMS from '@/stories/mocks/GALLERY_ITEMS';
 
+const FIGURE_SELECTOR = '.gallery__figure';
+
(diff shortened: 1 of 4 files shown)


drapisarda:
> I'm just a bit skeptical with the name `onCurrentItemClick`. Let's see what the others say!

I chosed it because we already have _currentItemIndex_ as data

tests/unit/components/HdGallery.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    });
+  });
+
+  it('renders component', () => {
+    expect(wrapper.html()).toMatchSnapshot();
+  });
+
+  it('switchs elements and emits the proper events clicks', () => {
+    const testIndex = 3;
+    const figure = wrapper.find('.gallery__figure');
```
ilyasmez: We can declare the selector on top of the file to make it clear that this test depends on it (the same way we are doing it in other tests):
```js
const FIGURE_SELECTOR = '.gallery__figure';
```

drapisarda: Done

PR #322, opened 2020-02-13
[T] Tests for HdSuggestionItem

drapisarda:


tests/unit/components/HdSuggestionItem.spec.js
```diff
@@ -0,0 +1,22 @@
+import { mount } from '@vue/test-utils';
+import HdSuggestionItem from '@/components/HdSuggestionItem.vue';
```
viniciuskneves: Wanna try using the import from `main.js` here? 😄

SinisaG:
Is this component used anywhere? We don't have a story for it and I did not find any reference of usage in other components. Maybe is a candidate for removal? @homeday-de/frontend-engineering ? 

If it's used, the match case is not tested. Please see here: https://coveralls.io/builds/28717266/source?filename=src/components/HdSuggestionItem.vue#L41 

When your inputted text matches one of the items, something should happen (highlight class). I'd wait to answer the question above though, before testing it further.

viniciuskneves:
> Is this component used anywhere? We don't have a story for it and I did not find any reference of usage in other components. Maybe is a candidate for removal? @homeday-de/frontend-engineering ?
> 
> If it's used, the match case is not tested. Please see here: https://coveralls.io/builds/28717266/source?filename=src/components/HdSuggestionItem.vue#L41
> 
> When your inputted text matches one of the items, something should happen (highlight class). I'd wait to answer the question above though, before testing it further.

Indeed, just checked it and we don't use this component internally. I'm 👍 on removing it!

aym3nb:
Yes, we don't use it in myHomeday too.

viniciuskneves:
@drapisarda could you create a PR removing this component?

PR #323, opened 2020-02-13
[T] Test for HdCalendar

drapisarda:


src/services/date.js
```diff
@@ -38,13 +38,13 @@ export const getIntlDateString = (locale, datesArray, localStringArgs) => datesA
...
+export const generateDateCycles = (cycleLengthWeeks, amountOfCycles, startDateOffset, startDate = new Date()) => {
   const WEEK_DAYS = 7;
   const cycleLengthDays = cycleLengthWeeks * WEEK_DAYS;
-  const startDate = getNDaysFromDate(new Date(), startDateOffset);
-  const endDate = getNDaysFromDate(new Date(), (cycleLengthDays * amountOfCycles) + startDateOffset);
+  const calculatedStartDate = getNDaysFromDate(startDate, startDateOffset);
+  const endDate = getNDaysFromDate(startDate, (cycleLengthDays * amountOfCycles) + startDateOffset);
 
-  return getDaysDateRange(startDate, endDate);
+  return getDaysDateRange(calculatedStartDate, endDate);
```
viniciuskneves: As you're expending it, can you also cover this scenario with tests? As `date service` is fully covered now (https://github.com/homeday-de/homeday-blocks/blob/develop/tests/unit/services/date.spec.js)

src/services/date.js
```diff
@@ -38,13 +38,13 @@ export const getIntlDateString = (locale, datesArray, localStringArgs) => datesA
   return accumulator;
 }, []);
 
-export const generateDateCycles = (cycleLengthWeeks, amountOfCycles, startDateOffset) => {
+export const generateDateCycles = (cycleLengthWeeks, amountOfCycles, startDateOffset, startDate = new Date()) => {
```
SinisaG: This is a good change. :+1:

src/services/date.js
drapisarda: Done!

PR #326, opened 2020-02-18
[T] Move services to root

viniciuskneves:
## Proposal

*For now you can ignore CI and so on*

- Moved services to root folder
  - **From**: \<rootDir\>/src/services
  - **To**: \<rootDir\>/services
- New way of importing services, so you don't need to fully import the service anymore
  - **From**
```javascript
import { DateService } from 'homeday-blocks';

const { generateDateCycles } = DateService; // or just use DateService.generateDateCycles
```
  - **To**
```javascript
import { generateDateCycles } from 'homeday-blocks/services/date';
```

### Advantage
- We can import just the function we need/want so we don't import the full service

### Disadvantage
- Some refactoring + breaking change
- [Mess(i)](https://sportbild.bild.de/fotos-skaliert/wird-gegen-inter-fehlen-barca-superstar-lionel-messi-201307831-66582448/2,w=993,c=0.sport.jpg) folder structure as `services` won't live inside `src` anymore
  - Regarding this we can think about moving other stuffs outside the `src` folder. We could have `homeday-blocks/components/form/HdInput` for example but that is not the goal of this PR (open for discussion though)

### DIFF
--- a/src/stories/HdCalendar.stories.js
+++ b/src/stories/HdCalendar.stories.js
@@ -1,8 +1,7 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import { storiesOf } from '@storybook/vue';
 import { action } from '@storybook/addon-actions';
-import { generateDateCycles } from 'homeday-blocks/src/services/date';
-
+import { generateDateCycles } from 'homeday-blocks/services/date';
 import HdCalendar from 'homeday-blocks/src/components/HdCalendar.vue';
 
 storiesOf('HdCalendar', module)
--- a/tests/unit/services/date.spec.js
+++ b/tests/unit/services/date.spec.js
@@ -5,7 +5,7 @@ import {
   getDaysDateRange,
   getIntlDateString,
   generateDateCycles,
-} from '@/services/date';
+} from 'homeday-blocks/services/date';
 
 
 describe('Date service', () => {
--- a/services/README.md
+++ b/services/README.md
@@ -1,0 +1,221 @@
+# Services
+Homeday Blocks contains a number of services that we found ourself needing across projects. Some of them are used by Homeday Blocks components, others 
+are here to be consumed in projects directly.
+
+Index:
+- [Breakpoints service](#breakpoints) - Used for setting (and getting) the global project breakpoints.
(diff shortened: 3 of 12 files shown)


src/stories/HdCalendar.stories.js
```diff
@@ -1,8 +1,7 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import { storiesOf } from '@storybook/vue';
 import { action } from '@storybook/addon-actions';
-import { generateDateCycles } from 'homeday-blocks/src/services/date';
-
+import { generateDateCycles } from 'homeday-blocks/services/date';
```
viniciuskneves: Take a look in here.

tests/unit/services/date.spec.js
```diff
@@ -5,7 +5,7 @@ import {
   getDaysDateRange,
   getIntlDateString,
   generateDateCycles,
-} from '@/services/date';
+} from 'homeday-blocks/services/date';
```
viniciuskneves: Take a look in here.

SinisaG:
I'm not sure advantages outweigh disadvantages in this case. I am specifically reserved about breaking the folder structure. 

I think services are not so big that they would be a serious overhead. We could instead:

- Rethink if services are grouped correctly? Can we break them down further to avoid potential bloating? 
- You can still access part of the service, just instead of doing `import { generateDateCycles } from 'homeday-blocks/services/date';`, you have `src` there as well. As this is "not usual", I think that is ok. So side affect here is smaller, in my opinion, then sacrificing folder structure of the project. More our library grows, more the folder structure will be important to keep it sane.

Open for discussion on this though.

viniciuskneves:
@homeday-de/frontend-engineering I would appreciate your input in here.

aym3nb:
In my opinion, I think even the advantage itself is a disadvantage because we need to know the absolute path of the service in order to import it. so I think the current way is cleaner.

viniciuskneves:
> Just an idea.
> What about exporting everything on main.js and then have `import { generateDateCycles } from 'homeday-blocks';` ?

I think it is too broad this way. I think we won't move this one forward. Gonna close it for now.

PR #331, opened 2020-02-21
[B] Fix HdGallery & HdGalleryCarousel 

aym3nb:
https://homeday.atlassian.net/browse/FET-133

This is the actual fix of this bug, after trying already in #327 

Would appreciate some input here @homeday-de/frontend-engineering

### DIFF
--- a/src/components/gallery/HdGalleryCarousel.vue
+++ b/src/components/gallery/HdGalleryCarousel.vue
@@ -60,6 +60,8 @@
 import Flickity from 'vue-flickity';
 import onResize from 'homeday-blocks/src/services/on-resize';
 import HdPager from 'homeday-blocks/src/components/HdPager.vue';
+import { mediaMatches } from 'homeday-blocks/src/services/breakpoints';
+import sassVariables from 'homeday-blocks/src/styles/variablesForJS.scss';
 
 export default {
   name: 'HdGalleryCarousel',
@@ -150,7 +152,10 @@ export default {
     },
     updateCurrentIndex(itemIndex) {
       this.currentIndex = itemIndex;
-      if (window.matchMedia('min-width: $break-tablet')) {
+
+      const mediaQuery = `(max-width: ${sassVariables.breakTablet})`;
+
+      if (mediaMatches(mediaQuery)) {
         this.$emit('input', itemIndex);
       }
     },
--- a/src/components/gallery/HdGalleryCarousel.vue
+++ b/src/components/gallery/HdGalleryCarousel.vue
@@ -153,9 +153,12 @@ export default {
     updateCurrentIndex(itemIndex) {
       this.currentIndex = itemIndex;
 
-      const mediaQuery = `(max-width: ${sassVariables.breakTablet})`;
+      const tabletMediaQuery = `(max-width: ${sassVariables.breakTablet})`;
 
-      if (mediaMatches(mediaQuery)) {
+      if (mediaMatches(tabletMediaQuery)) {
+        // Should not emit on desktop, because Flickity is used as thumbnails carousel there
+        // Whereas on mobile we use it as a gallery, so we should emit every time slide is changed.
+        // Prevents this bug https://homeday.atlassian.net/browse/CE-1025
         this.$emit('input', itemIndex);
       }
     },
--- a/src/styles/variablesForJS.scss
+++ b/src/styles/variablesForJS.scss
@@ -1,0 +1,8 @@
+@import './variables';
+
+:export {
+  breakMobile : $break-mobile;
+  breakTablet : $break-tablet;
+  breakDesktop : $break-desktop;
+  breakDesktopWide : $break-desktop-wide;
+}
--- a/src/styles/_variables.scss
+++ b/src/styles/_variables.scss
@@ -1,3 +1,5 @@
+@import './colors';
+
 $default-border-radius: 2px;
 
 /* FONTS */


src/components/gallery/HdGalleryCarousel.vue
```diff
@@ -150,7 +152,10 @@ export default {
     },
     updateCurrentIndex(itemIndex) {
       this.currentIndex = itemIndex;
-      if (window.matchMedia('min-width: $break-tablet')) {
+
+      const mediaQuery = `(max-width: ${sassVariables.breakTablet})`;
+
+      if (mediaMatches(mediaQuery)) {
```
SinisaG: @aym3nb pin pointed the bug here, the problem is that we use the same instance of flickity on desktop and mobile. However flickity works differently on those devices (on desktop index is the number of the page, while on mobile, it's the index of the image active). So on desktop, we should not emit this event. 

@aym3nb , please add more context here, if needed.

src/styles/variablesForJS.scss
```diff
@@ -0,0 +1,8 @@
+@import './variables';
+
+:export {
+  breakMobile : $break-mobile;
+  breakTablet : $break-tablet;
+  breakDesktop : $break-desktop;
+  breakDesktopWide : $break-desktop-wide;
+}
```
SinisaG: @homeday-de/frontend-engineering : Any potential issue you see here, whit exposing those variables and using them?

drapisarda: Looks good to me.

viniciuskneves: I think it is useful, sliders (https://github.com/homeday-de/sliders/blob/develop/src/styles/variables.scss#L101) also uses it.

aym3nb:
@viniciuskneves thanks for the review, for examples and details check https://homeday.atlassian.net/browse/FET-133

src/styles/variablesForJS.scss
SinisaG: We remembered you've demoed this with sliders - and used it! :)

src/components/gallery/HdGalleryCarousel.vue
```diff
@@ -150,7 +152,10 @@ export default {
     },
     updateCurrentIndex(itemIndex) {
       this.currentIndex = itemIndex;
-      if (window.matchMedia('min-width: $break-tablet')) {
+
+      const mediaQuery = `(max-width: ${sassVariables.breakTablet})`;
```
mlande: Only thing I'd suggest here, for the sake of documentation, is naming this variable more specifically, and adding a comment next to the if condition which explains the need for this "magical" condition.  Something like:

```
const tabletMediaQuery = `(max-width: ${sassVariables.breakTablet})`;

// Should not emit on desktop, as Flickity behaves differently there. 
// On desktop, the index represents the page number, not the active image. See [link to docs here?]. 
// Prevents this bug [link to bug ticket here?] 
if (mediaMatches(tabletMediaQuery)) { ... }
```

Thoughts?

aym3nb: thanks for the review and the suggestions, I totally agree 👌

viniciuskneves:
> https://homeday.atlassian.net/browse/FET-133

Yes, I've checked the ticket itself but I was suggesting something more code/debug wise instead of the bug video so it would be better to understand the bug code-wise as well.

src/styles/variablesForJS.scss
ilyasmez: The solution above won't work cross projects, because these variables are meant to be overwritten by each project, to match the needed breakpoints. So `sassVariables.breakTablet` will always be `768px` (as defined [here](https://github.com/homeday-de/homeday-blocks/blob/f0cee4815e333cfd15cba716b9a1354eea7a5a03/src/styles/_variables.scss#L50
)), while in myHomeday for example it should be `600px`.

I suggest using the breakpoints service for this, and defining the breakpoints globally on each project.
Or passing the breakpoint as a prop (ad having `sassVariables.breakTablet` as default).

SinisaG: Yeah, that's a concern. @aym3nb can you create an update ticket for this one? So we address it better. We can pass it as prop, would make it more robust and inline with our breakpoints strategy (with default value). :+1:

SinisaG: Actually, we declare variables as `default` in our scss - https://github.com/homeday-de/homeday-blocks/blob/develop/src/styles/_variables.scss#L49 
So if they are overwritten in the project, they would be picked up correctly during the build here as well, wouldn't they? Otherwise breakpoints of the project would not be inline with breakpoints here. 
Anyway, I guess my point is, I need to try it, to wrap my head properly around it. :) As this is already used in myHomeday, perhaps we can quickly check it there, if correct value is picked up or not?

ilyasmez: My guess is that they wouldn't, because we are getting the variables directly from the `_variables.scss` file, and not from where they are overwritten.
Yeah, let's try it and see.

aym3nb:
@ilyasmez yes I already thought about that point, I'll add it later as an improvement.  
And the reason why I had to release is that Sinisa and I already did an urgent release a week ago [https://github.com/homeday-de/homeday-blocks/tree/v5.0.1](/homeday-de/homeday-blocks/tree/v5.0.1) and that was before the new alias and colors.

src/styles/variablesForJS.scss
SinisaG: Those values would be set only if you did not define them. Have a look here: https://sass-lang.com/documentation/variables#default-values

> To make this possible, Sass provides the !default flag. This assigns a value to a variable only if that variable isn’t defined or its value is null. Otherwise, the existing value will be used.

So I think it could work. I imagine we inject our css variables at beginning with webpack, then the rest is loaded. Agree, it's to abstract for me to be certain at this point, so let's try it. :)

SinisaG: Just quickly checked the build of [myHomeday](https://my.homeday.de/js/app.85400a0c.js). 

```
{breakMobile:"460px",breakTablet:"600px",breakDesktop:"900px",breakDesktopWide:"1200p"}
```

Looks good. :ok_hand:

ilyasmez: I was shocked to see that it works too.

I did some tests and found out that maybe it works because we are still using the auto scss variables import in myHomeday. I assume that line gets injected in the beginning of `variablesForJS.scss` as well.

Snipped from vue.config.js:
```js
module.exports = {
...
css: {
    loaderOptions: {
      sass: {
        prependData: `
          @import "@/styles/variables.scss";
        `,
      },
    },
  },
};
```
Which means that even the `@import './variables';` is not needed in that case and you'll have access to all the projects variables (for example `$sidebar-width` which is myHomeday specific).

**tl;dr**
It will work just for myHomeday and not the other projects.

PR #347, opened 2020-02-27
[F] Default breakpoints for the responsive service

ilyasmez:
https://homeday.atlassian.net/browse/FET-146

While working on this one I noticed that we might want to discuss breakpoints in our upcoming DS meeting.
Because it seems to me like they weren't set with "mobile-first" in mind.

For example with this:
```
$break-mobile: 460px !default;
$break-tablet: 768px !default;
```
**is `500px` mobile or not?**
if it is, then we should probably get rid of `$break-mobile` and consider anything under `768px` to be mobile.
if it's not, then we should adapt the naming from `$break-mobile` to `$break-tablet`, `$break-tablet` to `$break-desktop`... and so on. (this is the approach I took in the responsive service breakpoints, that's why I ended up with `desktopExtraWide`).

`$break-mobile` doesn't make sense in mobile-first projects.

Check how we are doing it in [customer-app](https://github.com/homeday-de/customer-app/blob/develop/src/styles/breakpoints.scss) and [realtor-app](https://github.com/homeday-de/realtor-app/blob/develop/src/styles/breakpoints.scss).

### DIFF
--- a/src/services/breakpoints.js
+++ b/src/services/breakpoints.js
@@ -1,26 +1,36 @@
-let breakpoints = {};
+import sassVariables from 'homeday-blocks/src/styles/variablesForJS.scss';
 
-export const matchMediaAvailable = (
-  typeof window !== 'undefined'
-  && typeof window.matchMedia === 'function'
-);
+let breakpoints = {
+  mobile: `(max-width: ${sassVariables.breakMobile - 1}px)`,
+  tablet: `(min-width: ${sassVariables.breakMobile}px)`,
+  desktop: `(min-width: ${sassVariables.breakTablet}px)`,
+  desktopWide: `(min-width: ${sassVariables.breakDesktop}px)`,
+  desktopExtraWide: `(min-width: ${sassVariables.breakDesktopWide}px)`,
+};
 
-export const mediaMatches = (breakpoint, {
-  matchMedia = matchMediaAvailable ? window.matchMedia : () => false,
-} = {}) => {
+export const matchMediaAvailable = window && typeof window.matchMedia === 'function';
+
+const defaultMatchMedia = matchMediaAvailable ? window.matchMedia : () => false;
+
+export function mediaMatches(
+  breakpoint,
+  { matchMedia = defaultMatchMedia } = {},
+) {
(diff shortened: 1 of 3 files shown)


ilyasmez:
@viniciuskneves Definitely! I just want to hear your ideas first. I'll document this and add test when we all agree on something.

aym3nb:
Why don't we use something more common as the default breakpoints? 

|Framework|Small|Medium|Large|Exra large|
|--- |--- |--- |--- |--- |
|Bulma|-|min: 769px  ("mobile")|min: 1024px  ("desktop")|min: 1216px  ("fullhd")|
|Bootstrap 4|min: 576px|min: 768px|min: 992px|min: 1200px|
|Tailwind|min: 576px|min: 768px|min: 992px|min: 1200px|
|Zurb Foundation|-|min: 640px|min: 1024px|min: 1200px|

SinisaG:
Agree we should discuss it and thanks for bringing it up. I've added to agenda for next sync (and tagged you @ilyasmez ). :+1:

ilyasmez:
@aym3nb thanks for sharing! I'm with the idea. We can even mix them up and come up with something that we can easily migrate to (e.g. `640px`, `992px`, `1200px`).

@SinisaG Thank you! I'll add more details to the agenda later, so everybody comes prepared, to hopefully end up with a decision instead of an action item.

src/services/breakpoints.js
```diff
@@ -1,26 +1,36 @@
-let breakpoints = {};
+import sassVariables from 'homeday-blocks/src/styles/variablesForJS.scss';
 
-export const matchMediaAvailable = (
-  typeof window !== 'undefined'
-  && typeof window.matchMedia === 'function'
-);
+let breakpoints = {
+  mobile: `(max-width: ${sassVariables.breakMobile - 1}px)`,
```
drapisarda, line 4: So, would we like to remove this one?

ilyasmez: Yes. And this PR is `on hold`.
There is a ticket to update the breakpoints:
https://homeday.atlassian.net/browse/FET-156

I'll tackle it next sprint.

PR #348, opened 2020-02-28
[T] export missing circletoPath function

aym3nb:
People reported a bug about not being able to see the sign button on the DBO page. After I checked I found that **circletoPath function** can't be imported from homeday-blocks utilsService.


tests/unit/components/__snapshots__/HdCalendar.spec.js.snap
```diff
@@ -3,17 +3,17 @@
...
-      <div class="calendar__row__item calendar__day">Mon</div>
-      <div class="calendar__row__item calendar__day">Tue</div>
-      <div class="calendar__row__item calendar__day">Wed</div>
+      <div class="calendar__row__item calendar__day">Do</div>
+      <div class="calendar__row__item calendar__day">Fr</div>
+      <div class="calendar__row__item calendar__day">Sa</div>
+      <div class="calendar__row__item calendar__day">So</div>
+      <div class="calendar__row__item calendar__day">Mo</div>
+      <div class="calendar__row__item calendar__day">Di</div>
+      <div class="calendar__row__item calendar__day">Mi</div>
```
viniciuskneves: Revert those changes as it is breaking CI (are you using german by default? hehehe)

viniciuskneves:
@aym3nb could you fix conflicts so we can merge it?

PR #351, opened 2020-03-06
[T] Use HdInputPassword in HdDynamicForm and HdPasswordConfirm

ilyasmez:
https://homeday.atlassian.net/browse/FET-118

If you check these stories of [HdDynamicForm ](https://blocks.homeday.dev/?path=/story/form-hddynamicform--login)or [HdPasswordConfirm](https://blocks.homeday.dev/?path=/story/form-hdpasswordconfirm--simple) for example, you will notice the absence of the visibility toggle in the password inputs, that's because we are still using HdInput with type password there (which is not supported since #250).

After releasing this we should create tickets to start using this in our projects (login form of myHomeday and also realtor app)


src/components/form/HdDynamicForm.vue
```diff
@@ -52,12 +55,9 @@ export default {
     },
     lang: {
       type: String,
-      default: '',
+      default: undefined,
```
mlande: What do you think about setting the de-facto default `de` here on the form level, as it is happening under the hood in the input components anyway?

ilyasmez: I actually set it to `undefined` for the purpose of always falling to the components' default. The components can be used individually (without the form component), so I wanted to avoid having default values on different levels, and make it clear that it's the components' defaults that we are using here.
To be honest, I also don't like having the `'de'` string literal everywhere 😅  we should have it defined in one place and maybe make it configurable on the library level as well. ([ticket created](https://homeday.atlassian.net/browse/FET-148))

mlande: Ok, gotcha! That makes sense, and thanks for creating the ticket

ilyasmez: Thank you for the review! ❤️

PR #353, opened 2020-03-09
[F] HdDynamicForm more field column flexibility

mlande:
**The Context**
When configuring a dynamic form, one can pass an array as an item in order to group multiple fields in one line. Example:
```
export default function getConfigWith111Layout() {
  return [
    {...fieldA},
    {...fieldB},
    [
       {...fieldC1},
       {...fieldC2},
       {...fieldC3}
    ]
  ]
}
```

**The Problem**
However, these fields would always be flexed equally on the line (1-1-1), so one cannot create other layouts (e.g. 2-1-1 or 1-1-2). This limits our design implementation possibilities.

**The Solution**
By simply allowing one more level of array nesting in the `items` prop, we can allow to easily produce 2-1-1 and 1-1-2 layouts, by simply doing something like this:

```
export default function getConfigWith211Layout() {
  return [
    {...fieldA},
    {...fieldB},
    [
       {...fieldC1},
       [
         {...fieldC2},
         {...fieldC3}
       ]
    ]
  ]
}
```

**Up For Discussion**

- How does this fit in the upcoming grid system?

- Am I missing some horrible edge case here?

_I should note that this is a requirement for us to implement the layout for the new Profile Page on myHomeday https://homeday.atlassian.net/browse/CE-602_

### DIFF
--- a/src/components/form/HdDynamicForm.vue
+++ b/src/components/form/HdDynamicForm.vue
@@ -2,17 +2,37 @@
   <form class="dynamicForm" @submit.prevent="submit" novalidate>
     <slot name="before"/>
     <div v-for="(line, i) in lines" class="dynamicForm__line" :key="`line-${i}`">
+      <template v-for="(item, index) in getItemsArray(line)">
+        <div
+        v-if="Array.isArray(item)"
+        class="dynamicForm__line__item dynamicForm__line__item--with-subitems"
+        :key="`input-array-${index}`"
+        >
+          <component
+            v-for="subItem in item"
+            ref="fields"
+            :is="getComponent(subItem.type)"
+            :key="`input-${subItem.name}`"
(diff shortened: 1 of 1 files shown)


mlande:
**I should note that this is a requirement for us to implement the layout for the new Profile Page on myHomeday https://homeday.atlassian.net/browse/CE-602**

mlande:
@ilyasmez that's a great suggestion, I think it may just work for us! Also, now I feel dumb for not thinking about it :)

@aym3nb what do you think? From a quick test, adding the style prop should be enough to implement the the profile UI, without the need to change blocks. Let's discuss this separately.

In the meantime I will put this PR on hold.

mlande:
Solved as suggested by Ilyas. Closing this

PR #354, opened 2020-03-11
[T] PEW 03/2020 - Move reusable components from myhomeday to homeday-blocks 

aym3nb:
### Added
- HdModal (#371)
- HdToggle (#372)
- HdDashedList (#373)

### Updated
- HdLoaderButton (#374)

PS: This PR was split into different PRs so It will be easier to review each component separately without blocking.

### DIFF
--- a/src/components/buttons/HdLoaderButton.vue
+++ b/src/components/buttons/HdLoaderButton.vue
@@ -1,41 +1,71 @@
 <template>
-  <div class="loaderButton" :class="progressClass">
-      <button
-        class="btn btn--primary loaderButton__button"
-        :style="buttonStyles"
-        ref="button"
-        @click="clicked"
-        @keyup.enter.space="clicked"
-        @transitionend="runTransitionQue"
+  <div
+    :class="loaderButtonClasses"
+    :style="{
+      height: `${buttonHeight}px`,
+    }"
+    class="loaderButton"
+    @keydown="setUsingMouse(false)"
+    @mousedown="setUsingMouse(true)"
+  >
+    <button
+      ref="button"
+      :style="buttonStyles"
+      class="btn btn--primary loaderButton__button"
+      @click="clicked"
+      @keyup.enter.space="clicked"
+      @transitionend="runTransitionQue"
+    >
+      <span class="loaderButton__button__label">{{ label }}</span>
+    </button>
+
+    <!-- Loading circle -->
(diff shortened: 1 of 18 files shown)


viniciuskneves:
@aym3nb could you split it into different PRs?
It will be a way easier to review them once they're isolated and we don't block a component if another one is pending/review requested.

src/components/buttons/HdLoaderButton.vue
```diff
@@ -131,23 +173,28 @@ export default {
     loadingState(state) {
       if (state === 'success') {
         this.loadedAmount = 1;
-      }
-      if (state === 'error') {
-        this.transitionsQue.push(this.setErrorState);
+      } else if (state === 'error') {
```
metal-gogo: Is there a particular reason to add the `else` here?
As `state === 'success'`, and `state === 'error'` are mutually exclusive, the `else` is not needed.
Or we want this to be extra clear?

src/components/buttons/HdLoaderButton.vue
```diff
@@ -158,18 +205,11 @@ export default {
...
-    runTransitionQue: debounce(function debouncedQueMethod() {
-      if (this.transitionsQue.length > 0) {
-        const nextInQueMethod = this.transitionsQue.shift();
-        nextInQueMethod();
-      }
-    }, 100),
     fakeLoading() {
       const bumpLoading = () => {
-        const stopTreshold = this.isStatic ? 1 : 0.90;
+        const stopTreshold = this.isStatic ? 1 : 0.79;
```
metal-gogo: What does that `0.79` stand for?
Could it be stored in a semantic named variable?

aym3nb:
@viniciuskneves  yes, you are right that would be better.
I'll split it into different PRs.

src/components/buttons/HdLoaderButton.vue
aym3nb: No particular reason, I just checked and It wasn't there before I think it was added by Prettier.

src/components/buttons/HdLoaderButton.vue
aym3nb: `0.79`  :  the stop threshold of fake loading transition ( 79% of the full circle )
yes, you are right it would be better to have it in a semantically named variable.

PR #356, opened 2020-03-12
[T] Add changelog action

SinisaG:
Testing ;)


SinisaG:
This does not work due to token not being shared cross forks. So it can't post the changelog here. It should work with local branch.

PR #367, opened 2020-03-16
[T] Improve test coverage for HdRadioButton

leandroinacio:
https://homeday.atlassian.net/browse/FET-57
Got 100% coverage and added output checks

### DIFF
--- a/tests/unit/components/buttons/HdRadioButton.spec.js
+++ b/tests/unit/components/buttons/HdRadioButton.spec.js
@@ -12,8 +12,7 @@ describe('HdRadioButton', () => {
   const value = 'ipsum';
   const label = 'dolor';
   const ICON_SELECTOR = '.radioButton__icon';
-  const BACKGROUND_STYLE = 'background-image: url(desktop-icon-url);';
-  const EVENT_EMITTED = 'select';
+  const ICON_HOVER_SELECTOR = '.radioButton__icon--isHover';
 
   beforeEach(() => {
     wrapper = mount(HdRadioButton, {
@@ -30,10 +29,15 @@ describe('HdRadioButton', () => {
 
   test('the component is rendered', () => {
     expect(wrapper.html()).toMatchSnapshot();
-    expect(wrapper.find(ICON_SELECTOR).attributes().style).toBe(BACKGROUND_STYLE);
+  });
+
+  test('shows the provided icon', () => {
+    expect(wrapper.find(ICON_SELECTOR).attributes().style).toEqual(expect.stringContaining(desktopIcon));
+    expect(wrapper.find(ICON_HOVER_SELECTOR).attributes().style).toEqual(expect.stringContaining(desktopIconHover));
   });
 
   test('on change, the proper method is invoked', () => {
+    const EVENT_EMITTED = 'select';
     const input = wrapper.find('input');
     expect(input.attributes().name).toBe(name);
     expect(input.attributes().value).toBe(value);
--- a/tests/unit/components/buttons/HdRadioButton.spec.js
+++ b/tests/unit/components/buttons/HdRadioButton.spec.js
@@ -11,6 +11,9 @@ describe('HdRadioButton', () => {
   const name = 'lorem';
   const value = 'ipsum';
   const label = 'dolor';
+  const ICON_SELECTOR = '.radioButton__icon';
+  const BACKGROUND_STYLE = 'background-image: url(desktop-icon-url);';
+  const EVENT_EMITTED = 'select';
 
   beforeEach(() => {
     wrapper = mount(HdRadioButton, {
@@ -27,10 +30,7 @@ describe('HdRadioButton', () => {
 
   test('the component is rendered', () => {
     expect(wrapper.html()).toMatchSnapshot();
-  });
-
-  test('test the background-image of the radio icon', () => {
-    expect(wrapper.find('.radioButton__icon').attributes().style).toBe('background-image: url(desktop-icon-url);');
+    expect(wrapper.find(ICON_SELECTOR).attributes().style).toBe(BACKGROUND_STYLE);
   });
 
   test('on change, the proper method is invoked', () => {
@@ -40,9 +40,8 @@ describe('HdRadioButton', () => {
     expect(wrapper.find('label').text()).toBe(label);
 
     // Check events emitted before and after change
-    expect(wrapper.emitted('select')).toBeFalsy();
+    expect(wrapper.emitted(EVENT_EMITTED)).toBeFalsy();
     input.trigger('change');
-    expect(wrapper.emitted('select')).toBeTruthy();
-    expect(wrapper.emitted('select')[0][0]).toMatchObject({ value, name });
+    expect(wrapper.emitted(EVENT_EMITTED)[0][0]).toMatchObject({ value, name });
   });
 });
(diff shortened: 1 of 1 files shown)


tests/unit/components/buttons/HdRadioButton.spec.js
```diff
@@ -24,24 +23,26 @@ describe('HdRadioButton', () => {
...
     expect(input.attributes().value).toBe(value);
-
     expect(wrapper.find('label').text()).toBe(label);
 
+    // Check events emitted before and after change
+    expect(wrapper.emitted('select')).toBeFalsy();
     input.trigger('change');
-
-    expect(mockedSelect).toHaveBeenCalled();
+    expect(wrapper.emitted('select')).toBeTruthy();
```
viniciuskneves:
```suggestion
```

You can delete this line as the new assertion already covers it.

tests/unit/components/buttons/HdRadioButton.spec.js
```diff
@@ -24,24 +23,26 @@ describe('HdRadioButton', () => {
...
-    wrapper.setMethods({ select: mockedSelect });
   });
 
   test('the component is rendered', () => {
     expect(wrapper.html()).toMatchSnapshot();
   });
 
+  test('test the background-image of the radio icon', () => {
+    expect(wrapper.find('.radioButton__icon').attributes().style).toBe('background-image: url(desktop-icon-url);');
+  });
```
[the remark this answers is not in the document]
viniciuskneves: The `.snap` file is created once we call `matchSnapshot` and it keeps track of changes over component updates.

The downside is that it is not that descriptive (magic happening in the background).

You could add this assertion in the first test https://github.com/homeday-de/homeday-blocks/blob/develop/tests/unit/components/buttons/HdRadioButton.spec.js#L31-L33 to make it more descriptive.

tests/unit/components/buttons/HdRadioButton.spec.js
```diff
@@ -24,24 +23,26 @@ describe('HdRadioButton', () => {
...
-
-    wrapper.setMethods({ select: mockedSelect });
   });
 
   test('the component is rendered', () => {
     expect(wrapper.html()).toMatchSnapshot();
   });
 
+  test('test the background-image of the radio icon', () => {
+    expect(wrapper.find('.radioButton__icon').attributes().style).toBe('background-image: url(desktop-icon-url);');
```
ilyasmez: We tend to declare the selectors on top of the file to make it clear what elements the tests depend on and to make it easier to update, because class names are prone to change :)
Example: [HdAlert.spec.js#L4-L5](https://github.com/homeday-de/homeday-blocks/blob/develop/tests/unit/components/HdAlert.spec.js#L4-L5)

tests/unit/components/buttons/HdRadioButton.spec.js
leandroinacio: @ilyasmez but then the test wouldn't be 100% sure that we only emit the "select" event after a "change". It can look like it could have been fired after the component was created.
With both, we make sure this is triggered only after the change event.

ilyasmez: Ah my bad, sorry!! I overlooked this last line: 
```js
expect(wrapper.emitted('select')[0][0]).toMatchObject({ value, name });
```
My comment was for @viniciuskneves because I wasn't sure about the reason of removing that line (I didn't see the line above 🙈 ).

tests/unit/components/buttons/HdRadioButton.spec.js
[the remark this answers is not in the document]
leandroinacio: Done 👍

tests/unit/components/buttons/HdRadioButton.spec.js
leandroinacio: Done 👍

tests/unit/components/buttons/HdRadioButton.spec.js
leandroinacio: Done 👍

tests/unit/components/buttons/HdRadioButton.spec.js
```diff
@@ -11,7 +11,9 @@ describe('HdRadioButton', () => {
   const name = 'lorem';
   const value = 'ipsum';
   const label = 'dolor';
-  const mockedSelect = jest.fn();
+  const ICON_SELECTOR = '.radioButton__icon';
+  const BACKGROUND_STYLE = 'background-image: url(desktop-icon-url);';
```
ilyasmez:
```suggestion
  const BACKGROUND_STYLE = `background-image: url(${desktopIcon});`;
```

tests/unit/components/buttons/HdRadioButton.spec.js
```diff
@@ -24,24 +26,22 @@ describe('HdRadioButton', () => {
         mobileIcon,
       },
     });
-
-    wrapper.setMethods({ select: mockedSelect });
   });
 
   test('the component is rendered', () => {
     expect(wrapper.html()).toMatchSnapshot();
+    expect(wrapper.find(ICON_SELECTOR).attributes().style).toBe(BACKGROUND_STYLE);
```
ilyasmez: I believe that this one can be on its own test: "Shows the provided icon" or something like this.
And to be a bit flexible you can use `stringContaining` instead of `toBe` in case we add more styles.

tests/unit/components/buttons/HdRadioButton.spec.js
viniciuskneves: I strongly recommend updating to @ilyasmez instead of using hard-coded strings. We define this`desktop-icon-url` above.

tests/unit/components/buttons/HdRadioButton.spec.js
```diff
@@ -11,7 +11,9 @@ describe('HdRadioButton', () => {
   const name = 'lorem';
   const value = 'ipsum';
   const label = 'dolor';
-  const mockedSelect = jest.fn();
+  const ICON_SELECTOR = '.radioButton__icon';
+  const BACKGROUND_STYLE = 'background-image: url(desktop-icon-url);';
+  const EVENT_EMITTED = 'select';
```
viniciuskneves: This is unnecessary IMHO but not a big deal.

tests/unit/components/buttons/HdRadioButton.spec.js
viniciuskneves: Also just be aware that this https://github.com/homeday-de/homeday-blocks/blob/develop/src/components/buttons/HdRadioButton.vue#L23-L26 can also be tested the way you're doing. I do recommend adding it as well 😄

tests/unit/components/buttons/HdRadioButton.spec.js
leandroinacio: I have removed this variable, since I'm using "stringContaining" as suggested below.

tests/unit/components/buttons/HdRadioButton.spec.js
leandroinacio: I've moved the EVENT_EMITTED to the scope of the test using it, but kept the variable since I'm using the same string twice.

tests/unit/components/buttons/HdRadioButton.spec.js
leandroinacio: Done 👍

PR #370, opened 2020-03-17
[T] Add .nvmrc file

metal-gogo:
### Added
- `.nvmrc` to make sure all the team uses the same Node version.
- `scripts/setup.mac.sh` script to help bootstrap the project from a Mac computer.
- `scripts/setup.ubuntu.sh` script to help bootstrap the project from a Ubuntu computer.

---

#### `nvm` explanation

`nvm` is a great tool for any Node.js developer. It enables a concern-free installation and easy switching between different versions, saving time for what matters.

This is useful especially for new team members, that may install a brand new version of Node, while the project was developed with an older version of Node. This sometimes brings some incompatibilities and could break the app.

#### Setup scripts explanation

As `nvm` is not as trivial to install, it's better to automate the process to install it on the project.

--- 

If you need want to make sure that you change to the project node version automatically, you can add the following alias into the `~/.bashrc` or `~/.zshrc`:

```bash
alias cd="cdnvm(){ cd $1; if [[ -f .nvmrc && -s .nvmrc && -r .nvmrc ]]; then <.nvmrc nvm install; elif [[ $(nvm current) != $(nvm version default) ]]; then nvm use default; fi; };cdnvm"
```

---

#### References
- https://www.sitepoint.com/quick-tip-multiple-versions-node-nvm/
- https://dev.to/d4nyll/automatic-nvm-use-34ol


.nvmrc
```diff
@@ -0,0 +1 @@
+v12.16.1
```
viniciuskneves: It would be cool to match both: https://github.com/homeday-de/homeday-blocks/blob/develop/.travis.yml#L3

.nvmrc
```diff
@@ -0,0 +1 @@
+v10.19.0
```
viniciuskneves: Doing so you can remove this setting from `.travis.yml` file. Reference: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions-using-nvmrc

I would just recommend you to comment the line and add the link as a source of documentation for us in the future instead of just deleting it and wait until magic happens hehehe

aym3nb:
Well done 👏
Yes, this should have been done a long time ago 😂

README.md
```diff
@@ -17,13 +17,33 @@ $ git remote add upstream git@github.com:homeday-de/homeday-blocks.git
...
+Don't forget to setup the [deeper shell integration](https://github.com/nvm-sh/nvm#deeper-shell-integration) in your console to take full advantage of `nvm`. You can achieve this by adding the following alias into your `~/.bashrc`, or `~/.zshrc` file:
+
+```bash
+alias cd="cdnvm(){ cd $1; if [[ -f .nvmrc && -s .nvmrc && -r .nvmrc ]]; then <.nvmrc nvm install; elif [[ $(nvm current) != $(nvm version default) ]]; then nvm use default; fi; };cdnvm"
+```
 
 ## Project setup
 ```
 npm install
 ```
```
aym3nb: I think we can remove this since both scripts take care of it.

PR #372, opened 2020-03-19
[F] Add HdToggle

aym3nb:
Reusable component moved from myhomeday during PEW 03/2020.


src/components/HdToggle.vue
```diff
@@ -0,0 +1,243 @@
...
+import { OnResizeService } from 'homeday-blocks';
+
+export default {
+  name: 'HdToggle',
+  props: {
+    title: {
+      type: String,
+      default: '',
+    },
+    canBeToggled: {
```
SinisaG: Can you please add a test covering this use case as well? I think should be tested, as it's external API of the component.

src/components/HdToggle.vue
```diff
@@ -0,0 +1,243 @@
...
+      this.bodyHeight = this.getScrollHeight(this.$refs.body);
+    },
+    ensureBodyScrolledToTop() {
+      if (!this.$refs.body) {
+        return;
+      }
+
+      this.$refs.body.scrollTop = 0;
+    },
+    disableInternalFocus() {
```
SinisaG: Focus is not covered with tests, do you think it would make sense to cover it? If you decided not to, as this is internal behavior, I think that's valid as well. :+1:

aym3nb:
Thanks for the review @SinisaG 🙏
I agree about handling those cases which I didn't notice when I was sick 😂

src/components/HdToggle.vue
```diff
@@ -0,0 +1,251 @@
...
+
+      if (!$innerFocusableElements.length) {
+        return;
+      }
+
+      Array.from($innerFocusableElements).forEach((el) => {
+        const currentTabIndex = el.getAttribute('tabindex');
+
+        if (currentTabIndex !== null) {
+          el.setAttribute('data-hd-tabindex-backup', currentTabIndex);
```
ilyasmez, line 150: `'data-hd-tabindex-backup'` can go in a constant as it's used in two other places.

src/components/HdToggle.vue
```diff
@@ -0,0 +1,251 @@
...
+      }
+
+      Array.from($innerFocusableElements).forEach((el) => {
+        const oldTabIndex = el.getAttribute('data-hd-tabindex-backup');
+
+        if (oldTabIndex !== null) {
+          el.setAttribute('tabindex', oldTabIndex);
+          el.removeAttribute('data-hd-tabindex-backup');
+        } else {
+          el.removeAttribute('tabindex');
```
ilyasmez, line 178: Here you might end up removing the a `tabidex` that wasn't set by this component.
I'd suggest that in this method instead of querying this `'a, button, [tabindex]'` query `'[data-hd-tabindex-backup]'`, so this way we get just the elements we changed, and the logic will be also simplified as you know that all the elements you get have this attribute (so you just set the old and remove the backup).

PR #374, opened 2020-03-19
[T] Update HdLoaderButton

aym3nb:
Updated from myhomeday during PEW 03/2020.


src/components/buttons/HdLoaderButton.vue
```diff
@@ -76,15 +109,23 @@ export default {
...
+      type: Boolean,
+      default: false,
+    },
   },
   data() {
     return {
       state: 'idle',
       loadedAmount: 0,
       transitionsQue: [],
+      runTransitionQue: debounce(() => {
```
metal-gogo: I believe that it should be `runTransitionQueue` :).

src/components/buttons/HdLoaderButton.vue
```diff
@@ -76,15 +109,23 @@ export default {
...
-  mounted() {
-    this.undrawAll();
+    disabled: {
+      type: Boolean,
+      default: false,
+    },
   },
   data() {
     return {
       state: 'idle',
```
viniciuskneves: I would take the advantage to make an "Enum" out of it. There are a bunch of String comparisons happening and it might get lost in the future.

src/components/buttons/HdLoaderButton.vue
```diff
@@ -182,10 +231,26 @@ export default {
     },
     clicked() {
       this.$emit('click');
-      if (this.state === 'idle') {
-        this.state = 'loading';
-        this.transitionsQue.push(this.fakeLoading);
+
+      if (this.disabled) {
+        return;
+      }
```
viniciuskneves: Just wanna know if the idea of having it.  We want to keep the button's state but we want to make it disabled?

What I mean is, if we set the `<button>` to disabled, we would have the following state:
image

With this `disabled` prop it looks clickable but it doesn't do anything:
image

Is it the desired behavior?

aym3nb: Yes at the moment if it is disabled it'll be clickable but it won't start loading which is not UX friendly so I'll adjust it.

viniciuskneves: I mean, you migrated those components from MyHomeday, right? Does it make sense this use case there?

src/components/buttons/HdLoaderButton.vue
```diff
@@ -76,15 +115,23 @@ export default {
       type: Number,
       default: 4,
     },
-  },
-  mounted() {
-    this.undrawAll();
+    disabled: {
+      type: Boolean,
+      default: false,
+    },
```
viniciuskneves: I would also cover this guy with tests as it has an impact on how this component behaves.

src/components/buttons/HdLoaderButton.vue
```diff
@@ -1,54 +1,93 @@
...
+      </g>
+    </svg>
   </div>
 </template>
 
 <script>
 import { getRandomInt, circleToPath } from 'homeday-blocks/src/services/utils';
 import debounce from 'lodash/debounce';
 
+const LOADING_STATE = Object.freeze({
```
viniciuskneves: 👍 

I would suggest you to export this constant so whenever someone imports it they can use the Enum as well.
Similar to what we've here:
- https://github.com/homeday-de/homeday-blocks/blob/develop/src/components/buttons/HdButton.vue (You can also `Object.freeze` this guy hehehe)
- https://github.com/homeday-de/homeday-blocks/blob/develop/main.js#L20

src/components/buttons/HdLoaderButton.vue
aym3nb: Yes, I did but so far I never used this component personally.
According to its usage in **myHomeday** this is the desired behavior.

src/components/buttons/HdLoaderButton.vue
aym3nb: Right, but there is a separate ticket for improving its tests.
https://homeday.atlassian.net/browse/FET-56

src/components/buttons/HdLoaderButton.vue
aym3nb: Well, that's how I implement an Enum in JS because what you meant is just a constant 😂

src/components/buttons/HdLoaderButton.vue
```diff
@@ -73,11 +73,11 @@
 import { getRandomInt, circleToPath } from 'homeday-blocks/src/services/utils';
 import debounce from 'lodash/debounce';
 
-const LOADING_STATE = Object.freeze({
```
viniciuskneves: You can keep the `Object.freeze` here hehehe

src/components/buttons/HdLoaderButton.vue
viniciuskneves: Yes but as you're adding things to it, I thing it is safe to cover those scenarios. The ticket mentions improving test coverage which would be "not yet covered scenarios".

src/components/buttons/HdLoaderButton.vue
aym3nb: no need a constant is fine too

src/components/buttons/HdLoaderButton.vue
aym3nb: Yeah right!

tests/unit/components/buttons/HdLoaderButton.spec.js
```diff
@@ -1,38 +1,85 @@
...
+    expect(wrapper.classes()).toContain(`${BASE_CLASS}--${LOADING_STATE.ERROR}`);
+
+    // at the end it should be back to idle if `resetOnSuccess` isn't set to false
+    wrapper.vm.setIdleState();
+    jest.runOnlyPendingTimers();
+    expect(wrapper.vm.state).toEqual(LOADING_STATE.IDLE);
+    expect(wrapper.classes()).toContain(`${BASE_CLASS}--${LOADING_STATE.IDLE}`);
+  });
+
+  it('shouldn\'t be disabled even if the prop `disabled` is set to true', () => {
```
viniciuskneves, line 74:
```suggestion
  it('should emit click when disabled', () => {
```

src/components/buttons/HdLoaderButton.vue
```diff
@@ -1,54 +1,96 @@
...
-      <!-- Loading circle -->
-      <svg class="loaderButton__svg loaderButton__svg--progressCircle" v-show="state == 'loading'" :width="`${buttonHeight}px`" :height="`${buttonHeight}px`" ref="circle">
-        <path
-          :d="circlePath"
-          :style="{strokeWidth: `${loadingCircleStrokeWidth}px`}"
-        />
-      </svg>
+    <!-- Loading circle -->
+    <svg
+      v-show="state == LOADING_STATE.LOADING"
```
viniciuskneves, line 24:
```suggestion
      v-show="isLoading"
```

You missed it here.

src/components/buttons/HdLoaderButton.vue
```diff
@@ -76,18 +118,30 @@ export default {
...
       loadedAmount: 0,
       transitionsQue: [],
+      runTransitionQueue: debounce(() => {
+        if (this.transitionsQue.length > 0) {
+          const nextInQueMethod = this.transitionsQue.shift();
+          nextInQueMethod();
+        }
+      }, 100),
+      isUsingMouse: false,
+      LOADING_STATE,
```
viniciuskneves, line 138:
```suggestion
```

You don't need it anymore

PR #375, opened 2020-03-19
[T] Write tests for HdAutocomplete

leandroinacio:
https://homeday.atlassian.net/browse/FET-35

97.5% coverage.
I wonder if this component is being used somewhere... note that:

1. "autocomplete" prop is a string on HdInput - had to change it to make sure it doesn't trigger a warning when  executing the tests
2. "onEnter" method, used on the enter keydown is referenced on the template, but never created 
3.  "selectSuggestion" is defined with a second optional parameter, but never used (this is why we have one line missing coverage)
4. we don't have a storybook for this component


SinisaG:
It does not look like it's being used. I think @OlegIvaniv moved it from Sliders, but was never adopted there. I quickly checked the project and there's no such component (not to be confused by google autocomplete). Do we use it anywhere else @homeday-de/frontend-engineering ?

If we don't use it, I would suggest to delete it (I know it's already been tested :grimacing: ).

tests/unit/components/form/HdAutocomplete.spec.js
```diff
@@ -0,0 +1,98 @@
...
+        lang: 'en',
+      },
+    });
+  });
+
+  test('the component is rendered', () => {
+    expect(wrapper.html()).toMatchSnapshot();
+  });
+
+  test('the lang prop validation accepts only danish and english', () => {
```
metal-gogo:
```suggestion
  test('the lang prop validation accepts only deutch and english', () => {
```

tests/unit/components/form/HdAutocomplete.spec.js
```diff
@@ -0,0 +1,98 @@
...
+    wrapper.trigger('keydown.down');
+    expect(wrapper.vm.$data.selectedSuggestion).toBe(OPTIONS[1]);
+    expect(wrapper.vm.currentSuggestionIndex).toBe(1);
+
+    wrapper.trigger('keydown.down');
+    expect(wrapper.vm.$data.selectedSuggestion).toBe(OPTIONS[2]);
+    expect(wrapper.vm.currentSuggestionIndex).toBe(2);
+
+    wrapper.trigger('keydown.down');
+    expect(wrapper.vm.$data.selectedSuggestion).toBeNull();
```
metal-gogo: I have a question about the UX of the suggestions.
On a plain HTML the `select` input normally gets to the last option and stays there.
Should we try to emulate that behavior?

tests/unit/components/form/HdAutocomplete.spec.js
```diff
@@ -0,0 +1,98 @@
...
+    wrapper.trigger('keydown.up');
+    expect(wrapper.vm.$data.selectedSuggestion).toBe(OPTIONS[1]);
+    expect(wrapper.vm.currentSuggestionIndex).toBe(1);
+
+    wrapper.trigger('keydown.up');
+    expect(wrapper.vm.$data.selectedSuggestion).toBe(OPTIONS[0]);
+    expect(wrapper.vm.currentSuggestionIndex).toBe(0);
+
+    wrapper.trigger('keydown.up');
+    expect(wrapper.vm.$data.selectedSuggestion).toBeNull();
```
metal-gogo: I have a question about the UX of the suggestions.
On a plain HTML the `select` input normally gets to the first option and stays there.
Should we try to emulate that behavior?

aym3nb:
Yes, I just checked in **myHomeday** and we never used it.

tests/unit/components/form/HdAutocomplete.spec.js
leandroinacio: Good question. The current code is just removing the selection from the suggestions.
Let's see what direction this component will take, if we keep it we will have to talk about what behavior is expected.

leandroinacio:
On sliders I can only see the HdGoogleAutoComplete.

ilyasmez:
As @SinisaG said, this component is not used anywhere (it was one of the very first test components).
Let's see if it makes sense to keep it or not. But I'm leaning more towards removing it, to avoid maintaining something we don't use. And sorry about your wasted efforts.
I think we should go through the ticket in our board and delete the ones that don't make sense anymore.

leandroinacio:
@ilyasmez not a wasted effort, was a good exercise for the first week :)
No problem at all, I just removed the component as suggested.

PR #378, opened 2020-03-23
[T] Write tests for HdPasswordConfirm

aym3nb:
https://homeday.atlassian.net/browse/FET-41


leandroinacio:
@aym3nb shouldn't we add a test to update passwordMain and passwordConfirm values, and assert the validation class/message?
And one for the strength meter, to make sure the feature is working.

aym3nb:
@leandroinacio thanks for the review 🙏
 Regarding passwordMain and passwordConfirm values, they are already tested in `HdInputPassword` tests.
And I agree about adding a test for the strength meter 😅

PR #379, opened 2020-03-24
[F] DS-Assets (mainly icons)

ilyasmez:
## DS-Assets (mainly icons)

I've finally managed to find some time to finalize the icons topic.

**This pull request includes:**
- A script to automate the process of adding or updating assets from Figma (you should add your access token to the .env.local file, I can DM you mine to test).
- A new component to inline icons (`HdIcon`).
- The new icons (small, big and bicolor).
- Using `HdIcon` and the new icons in the current components and removing the old ones.
    - The input icons are the main exception because they are being used as background and some projects depend on that as far as I remember (Sliders?), so I just removed the unused ones and put them in a seperate folder, but I can't wait to delete them! :D If they are still used somewhere, we have to add the behaviour to `HdInput` and inline icons instead of using them as background.
- Categorizing the stories (Introduction, Components, Directives, Assets).
- Stories to list the avaialble icons
- Tests and docs


src/assets-importer/README.md
```diff
@@ -0,0 +1,33 @@
+# Assets Importer
+
+### Prerequisites
```
SinisaG: :+1:

src/assets-importer/index.js
```diff
@@ -0,0 +1,129 @@
...
+  CONFIG.FIGMA_FILES.forEach(FIGMA_FILE => importAssetsFromFigma({
+    fileKey: FIGMA_FILE.KEY,
+    format: FIGMA_FILE.FORMAT,
+    filenameRegex: FIGMA_FILE.FILENAME_REGEX,
+    shouldMatchRegex: FIGMA_FILE.SHOULD_MATCH_REGEX,
+    dist: FIGMA_FILE.DIST,
+  }));
+};
+
+importAll();
```
SinisaG: You could move the call to the top, so as soon as we open the file, we see the main part (without scroll). I usually, when I use function definition, put main part on top (or exposed exposed members - is something I've picked years ago working on [Angular 1](https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#accessible-members-up-top)  actually :D ) and then who's/when interested you can scroll down to implementation details. 

However, we often assign function to variables (with `const` often being the preferred way) and due to hoisting we have to reverse the order. If it's like that due consistency, then all good also :+1:  

Since mentioned already, what are you preferred ways @homeday-de/frontend-engineering ?

SinisaG:
Oh yeah, when running it locally, 1 updated icons was already available (if you want to reimport before merging). :D

src/assets-importer/index.js
ilyasmez: Ah I see your reference, because use used "main" in your comment, C and Python came directly to my mind :D

I usually declare then use (as this rule forced me to do: [no-use-before-define](https://eslint.org/docs/rules/no-use-before-define) :D), but I see your point, it's better to know what's happening from the beginning.

This is something that bothered me a bit (the call inside the file), so I'll change it to the command line, I'll keep the file just for the definitions, and invoke the `importAll` (and maybe also add an `importFileByName` function as well, to avoid importing everything everytime) from the command.

Thanks for bring up this topic, it confirms my concerns :D

SinisaG: Ah, yes...no use before define. :) It's good to reason and have this discussions to see different approaches. :+1:   We can also keep it, I think it's a matter of preference. 

On John Papa (the Angular reference) note - he's really cool, has usually some great tips (even old angular guide gives some good reasoning for code structuring outside angular). :ok_hand:

src/assets-importer/index.js
```diff
@@ -0,0 +1,129 @@
...
+
+  createIndex({
+    files: downloadedFiles,
+    dist,
+    exportNameRegex: new RegExp(`ic_(.*).${format}`),
+    suffix: 'Icon',
+  });
+}
+
+function createIndex({ files, dist, exportNameRegex = '*', suffix = '' }) {
```
metal-gogo: I would suggest renaming this function into `createIndexFile` as I had to read the whole function to understand that it was referring to the index file to export all the assets.

ilyasmez: Done 👍

src/assets-importer/index.js
ilyasmez: I went with something similar, please check the new commits :)

PR #383, opened 2020-03-26
[T] Add missing checkbox + radiobutton state to HD Blocks

leandroinacio:
https://homeday.atlassian.net/browse/CA-2222

Also fixed a state change that was not being rendered on HdCheckbox, line 11.
Design should match this:
https://www.figma.com/file/KdznSjw9btqHEKdgmm2UT4co/Design-System-UI-Patterns?node-id=7%3A99


viniciuskneves:
@leandroinacio could you fix conflicts 🙈?

leandroinacio:
Fixed 👍

src/components/form/HdCheckbox.vue
```diff
@@ -34,13 +35,15 @@
...
       @blur="handleBlur"
     >
-      <div class="checkbox__inner__box">
-        <div class="checkbox__inner__box__overlay"></div>
-        <div class="checkbox__inner__box__border"></div>
-        <HdIcon
-          :src="checkIcon"
-          class="checkbox__inner__box__tick"
-        />
+      <div class="checkbox__inner__box--shadow">
```
ilyasmez, line 38: `--shadow` here is not a modifier. It's more like a wrapper or container.
But looking at the component, BEM is being used as BEE...EEM 😄  (this is how we were doing it some time ago).
If fixing how BEM is being used in this component is out of this PR's scope, I'd suggest to go with the flow and use `__shadow` for now.

leandroinacio: Makes sense. 
I have just changed the class name for now, maybe fix BEM should be addressed in a future ticket.

PR #385, opened 2020-03-27
[B] Fixing assets url on HdModal

leandroinacio:
Assets were returning error for projects using Homeday-blocks.

image

### DIFF
--- a/src/components/HdModal.vue
+++ b/src/components/HdModal.vue
@@ -170,7 +170,7 @@ export default {
     position: absolute;
     top: $stack-s;
     right: $inline-s;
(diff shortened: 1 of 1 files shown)


ilyasmez:
Hi Leandro,
I've already addressed this issue in the DS-Assets PR.
Because we won't use assets this way anymore, please check the PR.
The component has not been released yet, so if there is no urgency to merge this PR, please don't, as it will create unnecessary conflicts.

leandroinacio:
Closing PR since, as @ilyasmez stated, we already have a fix in place and being reviewed. 👍

PR #400, opened 2020-04-07
[T] Upgrade vue-test-utils

viniciuskneves:
This PR addresses: https://github.com/homeday-de/homeday-blocks/pull/232

I do also recommend you to read the official changelogs: https://github.com/vuejs/vue-test-utils/releases

---

All in all it tackles the following breaking change: https://github.com/vuejs/vue-test-utils/issues/1137

I've also removed some tests, refactored others and cleaned up a bit.

I've also introduced the same `wrapperFactoryBuilder` that we're using in Realtor App. Its API maps to `mount`/`shallowMount` and instead of using `propsData` you just use `props`. I'm creating the wrapper inside each test to avoid calling `await wrapper.vm.$nextTick()` a lot of times.


viniciuskneves:
Yes, coverage dropped =]

I can work on improving it in another PR if you agree.

tests/unit/components/HdLink.spec.js
```diff
@@ -32,9 +32,10 @@ describe('HdLink', () => {
...
-
-    wrapper.setProps({
-      to,
+    const wrapper = wrapperBuilder({
+      props: {
+        to,
+      },
     });
 
     expect(wrapper.is('router-link')).toBe(true);
```
aym3nb, line 41: The test passes but it is logging this warning : `[Vue warn]: Unknown custom element: <router-link>`
I would suggest to make use of  `RouterLinkStub` (https://github.com/vuejs/vue-test-utils/pull/366) to have a warning free test 😂✅
```
import { RouterLinkStub } from '@vue/test-utils';

...

it('should render <router-link> on :to', () => {
    const to = 'www.homeday.de';
    const wrapper = wrapperBuilder({
      props: {
        to,
      },
      stubs: {
        RouterLink: RouterLinkStub,
      },
    });
    expect(wrapper.find(RouterLinkStub).props().to).toBe(to);
  });
```

viniciuskneves: Yeah, I saw this one. Tried the stub but it didn't work. Gonna try it once again. Thanks for checking it.

I will also merge your latest tests here to keep it in sync =]

tests/unit/components/HdAlert.spec.js
```diff
@@ -1,54 +1,60 @@
...
     const TEST_TYPE = 'warning';
-
-    wrapper.setProps({
-      type: TEST_TYPE,
+    const wrapper = wrapperBuilder({
+      props: {
+        type: TEST_TYPE,
+      },
     });
+    // It doesn't make sense to go through all the possible types (we may have many)
```
metal-gogo, line 36: Maybe we could have an array of types, and assert that every type is working correctly?

viniciuskneves: I will update this test to match: https://github.com/homeday-de/homeday-blocks/blob/develop/tests/unit/components/HdLink.spec.js#L21

viniciuskneves:
@homeday-de/frontend-engineering May I proceed here?

viniciuskneves:
image

PR #401, opened 2020-04-07
[F] HdInputFormatter

viniciuskneves:
It adds `HdInputFormatter` to our library.

This component is a wrapper on top of `HdInput` which accepts a `formatter` function and once the `HdInput` is not focused anymore, the `formatter` function is applied in its value and is displayed.

Useful once we deal with currencies for example or if you want thousand separators to be displayed...

Keep in mind that the `formatter` is not applied once the input is focused so it doesn't work like a mask.


src/stories/HdInputFormatter.stories.js
```diff
@@ -0,0 +1,39 @@
...
+
+        return '';
+      },
+    },
+    watch: {
+      value(value) {
+        action('input')(value);
+      },
+    },
+  }), { percy: { skip: true } });
```
ilyasmez, line 39: You're skipping it because it's based on HdInput?

viniciuskneves: Yep, shall I add a comment? If so, could you make a `insert suggestion`? =]

ilyasmez: Here you go:
```suggestion
  }), {
    // Doesn't make sense to test the UI of this components because it's based on HdInput
    percy: { skip: true },
  });
```

PR #407, opened 2020-04-14
[T] Version bump: v8.0.1

leandroinacio:


ilyasmez:
@leandroinacio 
We have new functionality (e.g. HdToggle) so that would be a minor.
And we are also removing a component (HdAutocomplete), I know that none of our projects depend on it, but to stay sane as this is an open source project somebody might be using it, I'm for bumping the major.

PR #415, opened 2020-04-21
[B] Don't shrink HdAlert icons

ilyasmez:
@viniciuskneves discovered this bug:
hdalert icon bug

As you can see when the message goes to more than one line, the icon shrinks (it should stay 24px*24px). This PR fixes that.


viniciuskneves:
*Percy discovered this bug 😄

PR #418, opened 2020-04-23
[T] Don't try to publish to Github packages

ilyasmez:
I left this code hoping that one day will start supporting Github Packages, but now I don't see us doing it anytime soon, especially after Github acquiring NPM (let's see what happens), so this code is giving us nothing but a failed action.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,4 +1,4 @@
-name: Publish to NPM and Github Packages
+name: Publish to NPM
 
 on:
   release:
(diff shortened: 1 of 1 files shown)


viniciuskneves:
Do you want me to wait before the release?

ilyasmez:
> Do you want me to wait before the release?

You can go on with the release, I'll merge this one afterwards.

ilyasmez:
Or we can also merge it and use the patch you're about to release as a validation.

PR #422, opened 2020-04-28
[T] Allow HdRange labels to be <html>

viniciuskneves:
As a background: we need it in our squad as the labels should be side by side but actually break into specific parts of the word and not randomly.

From the tests, that is the scenario **before** this PR once an `<html>` tag was used:
image

Now it will look like expected:
image

I don't think we need to add or update anything in the stories. Let me know if you think we should update something there.

Another question: Shall we do the same for the [tooltip value](https://github.com/homeday-de/homeday-blocks/blob/develop/src/components/form/HdRange.vue#L53)?

### DIFF
--- a/src/components/form/HdRange.vue
+++ b/src/components/form/HdRange.vue
@@ -41,7 +41,7 @@
         <p
           v-if="labels[stepIndex]"
           class="range__step-label"
-          v-text="labels[stepIndex]"
+          v-html="labels[stepIndex]"
         />
       </button>
     </div>
--- a/tests/unit/components/form/HdRange.spec.js
+++ b/tests/unit/components/form/HdRange.spec.js
@@ -133,7 +133,7 @@ describe('HdRange', () => {
   });
 
   describe('Labels', () => {
-    const TEST_LABELS = ['label1', 'label2', 'label3', 'label4'];
+    const TEST_LABELS = ['label1', 'label<br>2', 'label3', 'label4'];
     const TEST_INDEX = Math.round(TEST_LABELS.length / 2);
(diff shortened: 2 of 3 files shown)


volcanioo:
I know this is a test but I want to say something about this part.

Should we use `<br>` in that case I don't know. Because we care about semantic HTML structure on our products. `<br>` usually using for break lines of address, poetry, or code samples. 

> Generally, the `<br>` element should only be used if the line break itself is an intrinsic part of the content. This might be the case, for example, with:
> 
> Read more: https://html.com/tags/br/#ixzz6KvsFHgk8

I mean, it's not a big deal but I'm just sharing my idea. 
We can separate that number like this too:

```html
<label data-breakline="2">Label</label>
```
```css
label:after {
      content: attr(data-breakline);
      display: block;
}
```

SinisaG:
Hey @viniciuskneves ,
is that an aesthetic break or is it because lack of space? 

Could CSS help solve those issues? There are some options to control the breaks and behavior. Please see here: https://css-tricks.com/where-lines-break-is-complicated-heres-all-the-related-css-and-html/

If this would help, I would prefer it, rather then rendering `html`, as you always have to consider where `html` is coming from and if it should be sanitized.

viniciuskneves:
> Hey @viniciuskneves ,
> is that an aesthetic break or is it because lack of space?

Aesthetic in our case. We have enough space to fit the text and so on but it looks a way better once it breaks into a specific part of the text.

The concrete example is
- Optimaler Startpreis aus statistischen Verfahren
- Optimaler Startpreis 
  aus statistischen Verfahren

But please, do not only focus on `<br>`, that was just an example.

PR #433, opened 2020-05-07
[B] Fix the IE11 support

ilyasmez:
The `vue-inline-svg` points to an uncomplied file in its package.json (the `module` property).
So I changed the import path to pick the compiled file. This is not ideal, because we depend on the location of the file (I made sure that we use an exact version), I'm planning to open an issue (or a PR) in their repo, so we can hand this better.

I've also updated the welcome page.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24596,9 +24596,9 @@
       "dev": true
     },
     "vue-inline-svg": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/vue-inline-svg/-/vue-inline-svg-1.2.0.tgz",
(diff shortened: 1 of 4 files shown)


ilyasmez:
@homeday-de/frontend-engineering 
I opened a [PR](https://github.com/shrpne/vue-inline-svg/pull/15) to fix the issue on the dependency level, it got merged and released. I switched to using the new version instead of doing it the hacky way (importing the file directly).
Please check the PR again.

That means that you don't have to transpile `vue-inline-svg` in your projects.
So this: https://github.com/homeday-de/customer-app/pull/1163
and this: https://github.com/homeday-de/kirby/pull/2086
can be reverted after upgrading to Blocks v11.0.1 and up.

PR #435, opened 2020-05-11
[B] Remove blurred transition of image slide on Gallery

volcanioo:
https://homeday.atlassian.net/browse/CA-2366

### DIFF
--- a/src/components/gallery/HdGalleryMedia.vue
+++ b/src/components/gallery/HdGalleryMedia.vue
@@ -119,6 +119,7 @@ export default {
     &__thumbnail {
       opacity: 0;
       transition: opacity ($time-s * 2) ease-in-out;
+      filter: blur(3px);
 
       &.isVisible {
         opacity: 1;
--- a/src/components/gallery/HdGalleryMedia.vue
+++ b/src/components/gallery/HdGalleryMedia.vue
@@ -119,12 +119,10 @@ export default {
     &__thumbnail {
       opacity: 0;
       transition: opacity ($time-s * 2) ease-in-out, transform .2s;
-      filter: blur(4px);
 
       &.isVisible {
         opacity: 1;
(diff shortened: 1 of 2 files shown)


volcanioo:
@ilyasmez We discussed the same ideas with Gregor. I already recommended that but he still wants to remove, let's talk about it together when he back to the office. 

At the same time, the blur effect needs to scale. We shouldn't remove the only scale, the reason for this: 
> I added a style too. Because image element borders did not blur when filter property works. I used the scale feature for the image. In that way, we can hide these unblurred corners when shows.
https://github.com/homeday-de/homeday-blocks/pull/423#issuecomment-622973589

About `transition`, I do agree, I'll update this.

ilyasmez:
I think that the blur is an essential part of the lazy loading.
Currently the sequence looks like this:
`[image 1] -(user clicks next)-> [blurred image 2 thumbnail] -(0.3s fade-out transition)-> [image 2]`
With your changes:
`[image 1] -(user clicks next)-> [pixelized image 2 thumbnail] -(0.3s fade-out transition)-> [image 2]`
For simplification, I excluded the scale animation in the example.

So as you can see it's not making it any better. Maybe the opposite.

If you want the user to have "a quick impression of the property", then I would suggest to decrease the `0.3s` to maybe `0.15s`. But showing them a pixelized photo won't help.

I got the reason behind the scale animation, but it causes an undesirable motion animation, so I'm not sure if the blurred borders are more noticeable. But this one is a trade-off.

volcanioo:
I do agree, I'll talk with Gregor about it and if he accepts that: I'll update the transition time and I'll remove the scale thing. Thank you for your cooperation @ilyasmez.

volcanioo:
I talked with Gregor, We decided to use the blurred effect for now.

PR #441, opened 2020-05-19
[F] RadioButton improvements

leandroinacio:
Changed the label to render html so we can add spans and customize the text.
Changed type to accept objects. That could be the case depending on the project (we have a case like that on sliders).
And the unit test was failing... not sure why, couldn't find any changes related to it...
for some reason the first $nextTick() didn't digest the event all the way before the assert.

### DIFF
--- a/src/components/buttons/HdRadioButton.vue
+++ b/src/components/buttons/HdRadioButton.vue
@@ -32,7 +32,7 @@
         class="radioButton__icon"
       />
     </div>
-    <label :for="name" class="radioButton__label" v-text="label" />
+    <label :for="name" class="radioButton__label" v-html="label" />
     <HdIcon
       :src="chevronIcon"
       class="radioButton__chevron"
--- a/src/components/buttons/HdRadioButton.vue
+++ b/src/components/buttons/HdRadioButton.vue
@@ -52,7 +52,7 @@ export default {
   props: {
     label: String,
     value: {
-      type: [String, Number],
+      type: [String, Number, Object],
       required: true,
     },
     name: {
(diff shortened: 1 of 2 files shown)


src/components/buttons/HdRadioButton.vue
```diff
@@ -52,7 +52,7 @@ export default {
   props: {
     label: String,
     value: {
-      type: [String, Number],
+      type: [String, Number, Object],
```
viniciuskneves, line 55: Could you link sliders PR here so we can better understand the use case?

tests/unit/components/notifications/HdNotifications.spec.js
```diff
@@ -43,6 +43,7 @@ describe('HdNotifications', () => {
     wrapper.setMethods({ getScrollHeight: mockedScrollHeight });
     wrapper.setProps({ notifications: NOTIFICATIONS.slice(-1) });
 
+    await wrapper.vm.$nextTick();
```
viniciuskneves, line 46: It is super weird. I just run the tests locally and it worked. Have you checked node version? I mean, is it getting the right version from nvm?

src/components/buttons/HdRadioButton.vue
leandroinacio: This PR, where the value I need is the option selected:
https://github.com/homeday-de/sliders/pull/1193
I could use the index of the option and then reference the option with the position at the array, but I don't see why not just pass the object.

tests/unit/components/notifications/HdNotifications.spec.js
leandroinacio: I'm some versions above, I'm trying to setup nvm on Windows and will update the results as soon as it works.

leandroinacio: That made the trick. I was two versions above and didn't see that before, thanks for the heads-up.

viniciuskneves: Cool! "Nice" to know that you've had some issues with new versions, we must keep 👀 once it is upgraded 😬

PR #442, opened 2020-05-22
[F] Add disabled state for HdLoaderButton

leandroinacio:
https://homeday.atlassian.net/browse/FET-176
-
Not sure why we had a test expecting to trigger the click when the button is disabled, but if there is a reason we should add some comments and a note on storybook.

### DIFF
--- a/src/components/buttons/HdLoaderButton.vue
+++ b/src/components/buttons/HdLoaderButton.vue
@@ -12,10 +12,10 @@
       ref="button"
       :style="buttonStyles"
       class="btn btn--primary loaderButton__button"
+      :class="customClasses"
       @click="clicked"
       @keyup.enter.space="clicked"
       @transitionend="runTransitionQueue"
-      :disabled="disabled"
     >
       <span class="loaderButton__button__label">{{ label }}</span>
     </button>
@@ -176,6 +176,11 @@ export default {
         borderWidth: setIfLoading(this.loadingCircleStrokeWidth),
       };
     },
+    customClasses() {
+      return {
+        'btn--primary-disabled': this.disabled,
+      };
+    },
   },
   watch: {
     loadedAmount(val) {
@@ -314,7 +319,7 @@ export default {
       pointer-events: none;
       width: 46px; /* make a circle */
       border-width: 4px;
-      border-color: getShade($quaternary-color, 50);;
+      border-color: getShade($quaternary-color, 50);
       border-style: solid;
       border-radius: 30px;
       background-color: transparent;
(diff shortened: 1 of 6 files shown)


aym3nb:
I thought about the same thing when I updated it from myHomeday.
But since the creator left and there are no notes I just figured the desired behavior from the previous usage in myHomeday.
So, I think we need to confirm with designers first.

viniciuskneves:
> @leandroinacio I thought about the same thing when I updated it from myHomeday #374
> But since the creator left and there are no notes I just figured the desired behavior from the previous usage in myHomeday.
> So, I think we need to confirm with designers first.

Check the comments here: https://github.com/homeday-de/homeday-blocks/pull/374#discussion_r400173384

src/components/buttons/HdLoaderButton.vue
```diff
@@ -15,6 +15,7 @@
       @click="clicked"
       @keyup.enter.space="clicked"
       @transitionend="runTransitionQueue"
+      :disabled="disabled"
```
viniciuskneves, line 18: If you're just forwarding the prop to the button, I would go to `v-bind="$attrs"` and remove it from the props definition. That is a minor and just a tip, as other button attributes would be automatically forwarded.

leandroinacio:
@aym3nb @viniciuskneves fyi talked to Greg, he said:

> Disabled = grey :heavy_check_mark:
Disabled has no hover state :heavy_check_mark:
Disabled “allows” the click in form of a ripple effect that triggers the form field (the button itself doesn’t change it’s color or type color) + the click requests if anything is properly filled (like on production)

So I'm going to work on these and add a note on storybook.

leandroinacio:
@homeday-de/frontend-engineering ready for another review round :)

tests/unit/components/buttons/HdLoaderButton.spec.js
```diff
@@ -105,4 +105,23 @@ describe('HdLoaderButton', () => {
...
+  it('should toggle disabled class', async () => {
+    // Enabled
+    const disabledClass = 'btn--primary-disabled';
+    const wrapper = wrapperBuilder({
+      props: {
+        disabled: false,
+      },
+    });
+    const button = wrapper.get(BUTTON_SELECTOR);
+    expect(button.attributes().class).not.toContain(disabledClass);
```
viniciuskneves, line 118: I think you could have used https://vue-test-utils.vuejs.org/api/wrapper/#classes

leandroinacio: Yeah makes sense, done 👍

PR #448, opened 2020-05-28
[F] Gitmoji CLI hook

viniciuskneves:
It introduces [Gitmoji CLI](https://github.com/carloscuesta/gitmoji-cli) as default once we're preparing our commit message so we don't need to type a commit message anymore.

Once you run `git commit` (you can do `-a` as well) it will fire [Gitmoji CLI](https://github.com/carloscuesta/gitmoji-cli) and you can follow the steps.

It has basically 3 steps:
- Emoji selection: You can type to search for an Emoji
- Commit title: It is what we usually see on GitHub
- Commit message: It is what is hidden once you click `[...]` on GitHub

~~By default Gitmoji CLI uses a param to know where to write the commit message which I'm setting to `COMMIT_EDITMSG`.~~
~~I've tried with some rebase and it is not the best experience as it tries to add a commit message to every commit squashed for example, but you can do it only once and cancel and edit the message at the end (default rebase step).~~

Please, check how it goes on Windows, I've no clue if it works there. 

---

You can follow the commits but I decided to install Husky as it allows us to forward `GIT_PARAMS` and Yorkie apparently doesn't (or it is not well documented at least). With `GIT_PARAMS` Gitmoji knows if it needs to write to `COMMIT_EDITMSG`, `MERGE_MSG` or `SQUASH_MSG` which makes the user experience a way better!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3831,6 +3831,12 @@
         "any-observable": "^0.3.0"
       }
     },
+    "@sindresorhus/is": {
+      "version": "0.14.0",
+      "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-0.14.0.tgz",
+      "integrity": "sha512-9NET910DNaIPngYnLLPeg+Ogzqsi9uM4mSboU5y6p8S5DzMTVEsJZrawi+BoDNUVBa2DhJqQYUFvMDfgU062LQ==",
+      "dev": true
+    },
     "@soda/friendly-errors-webpack-plugin": {
       "version": "1.7.1",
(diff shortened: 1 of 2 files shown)


aym3nb:
@SinisaG here is how it looks on Mac 😄  
image

SinisaG:
@aym3nb : Vau, looks a bit better. :smile:  Thanks for sharing! :+1:

ilyasmez:
Annotation 2020-05-29 173618
And this is how it looks like using WSL if you're curious :D

PR #452, opened 2020-06-15
[F] validate only non-empty input value

viniciuskneves:
Thread: https://homeday.slack.com/archives/C41SHFS5R/p1591690652006900

---

Right now if an input is empty (and not required) it is still validated.

See the GIF below:
- Empty state is not validate first ✅
- Invalid e-mail is validated and error message is shown ✅
- Valid e-mail is validated and error message is removed ✅
- Empty state is validated and error message is shown ❌
Example

The new behavior will validate empty inputs only if they're required, otherwise the validation method is not called.

### DIFF
--- a/src/components/form/HdInput.vue
+++ b/src/components/form/HdInput.vue
@@ -194,6 +194,8 @@ export default {
     showError(errorMessage) {
       this.error = errorMessage;
       this.isValid = false;
+
+      return false;
     },
     showHelper(message) {
       this.helper = message;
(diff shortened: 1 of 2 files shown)


src/components/form/HdInput.vue
```diff
@@ -205,12 +205,16 @@ export default {
...
+      } else if (!this.isEmpty) {
+        if (this.currentType === 'email' && !validateEmail(this.value)) {
+          this.showError(this.t.FORM.VALIDATION.INVALID_EMAIL);
+        } else if (this.currentType === 'date' && !validateDate(this.value)) {
+          this.showError(this.t.FORM.VALIDATION.INVALID_DATE);
+        } else if (this.customRules.length) {
+          this.validateCustomRules();
+        } else {
+          this.hideError();
+        }
```
drapisarda, line 217: What about putting some returns and remove all those else's?

viniciuskneves: https://github.com/homeday-de/homeday-blocks/pull/452/files#diff-2c5cff5f7c1fb0874035f6dc78e56bb4R222

leandroinacio: How about an array of objects mapping field fieldType, validationFunction, errorMessage? Like:
```
 [
{
  fieldType: 'email',
  validationFunction: validateEmail,
 errorMessage: this.t.FORM.VALIDATION.INVALID_EMAIL
}
]
```

This way we can map every validation we have in a single file (you can export it from another file) and as it grows, there is no need to add if/else. You can do a .find(), call the validation if one validation fails.
Only one showError()/hideError() would be made.

leandroinacio: You can even previously .filter() the field to get the validations matching the type of the field you are currently working with.
Or you can change the structure to be like this, and access the fieldType by key:
{
  'email': [
    {
       fieldType: 'email',
       validationFunction: validateEmail,
      errorMessage: this.t.FORM.VALIDATION.INVALID_EMAIL
    }
  ],
  'text': [...]
}

viniciuskneves: @leandroinacio @drapisarda you can check my last commit, I've tried to remove some `else` clauses.

The idea from the validator map is nice but I don't see us updating it that much to justify the effort of implementing it at the moment. I can see it inside a project with more complex business rules or move it here if we "extend" this component with our "base" business rules.
If you want we can bring this topic for a further discussion about how we deal with validations and approach it differently.
Feel free to bring to team sync if you think it makes sense.

leandroinacio: I will add it to the team sync if you don't mind, cause I will need to implement async validations for sliders, so I will be touching this code eventually.
Code looks good to me 👍

viniciuskneves: Ohhhh, nice one! Please, bring it to the table as it might bite you later on!

viniciuskneves:
@drapisarda ?

PR #461, opened 2020-06-19
[B] Allow decimal numbers on webkit browsers

ilyasmez:
### The bug
Currently if you try to type a decimal number in an HdInput with type "number" you won't be able to do so on webkit browsers (tested on Safari mobile and desktop and Chrome for iOS).
The reason is that when you type a dot to add  the decimal part, webkit browsers emit an `input` event, which is not the case for the modern browsers (take that 🍎!) and will make use parse the new value, so `123.` will immediately become `123`, so the user won't even see the dot.

### Suggested solution
Before we emit the new value we compare the parsed number if it's the same as the current value, if they are, we don't emit the custom `input` event. This way, the parent won't know about the added dot, but when we add a number after the dot the parsed value will change and that's when we emit the `input` event.
**Example:**
| Input value | Emitted value | Model value |
|-------------|---------------|-------------|
| `"123"`     | `123`         | `123`       |
| `"123."`    | -             | `123`       |
| `"123.4"`   | `123.4`       | `123.4`     |

### DIFF
--- a/src/components/form/HdInput.vue
+++ b/src/components/form/HdInput.vue
@@ -178,8 +178,9 @@ export default {
       if (this.currentType === 'number' && newValue !== '') {
         newValue = parseFloat(newValue);
 
-        // If the parse value has no change we don't emit the new value
-        // because it might suppress the the decimals' dot
+        // If the parsed value is the same as the current value
+        // we don't emit the custom input event
+        // because it might suppress the the decimal deparator on webkit browsers
         if (newValue === this.value) {
           return;
         }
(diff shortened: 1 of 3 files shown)


ilyasmez:
> I would just ask you to add a test for this case.

What do you want to test exactly? Because our tests are using Jsdom, so the results will always be the same.
This is related to the html specs (this [one ](https://html.spec.whatwg.org/multipage/input.html#number-state-(type=number))to be specific) that webkit is not fully respecting or using an older version of.
 
> If someone wants to try: https://codepen.io/viniciuskneves/pen/zYrNWQL --> I've created this CodePen where the on `input` I `console.log` the value. If you have a Mac you can see that it emits `12.` instead of ` `.

Thanks for checking it out, I forgot to share mine, where I did most of the debugging :D

ilyasmez:
@viniciuskneves I'll merge this one, but we can continue the conversation here and I'll follow up on it if necessary.

PR #465, opened 2020-06-22
[F] ✨ HdTileSelect component

viniciuskneves:
https://homeday.atlassian.net/browse/FET-162

---

It is a migration of the following component: https://github.com/homeday-de/realtor-app/blob/develop/src/components/HdTileSelect.vue

I tried to simplify a bit the API and remove unused props like: `name` and `required`.

I've also taken the chance to remove the custom `rows` prop to leave the component responsive by its own (it wasn't before).

The `items` prop now accepts just an Array of values as well so you don't need to define `text` and `value` always the same when using the component.

I've had issues testing `HdTileSelectEditableItem` because of the `v-if` logic. The test didn't run once `click` happened so I went for a forced approach.

Take your time to review it as I will also take a time to test it in Realtor App and see it everything works as expected. Customer squad can also play with it as they're using it on their side as well.


src/components/HdTileSelectItem.vue
```diff
@@ -0,0 +1,79 @@
...
+    type="button"
+    :class="{
+      'tile-select-item--selected': isSelected,
+    }"
+    @click="emitInput"
+    @keydown.enter.space.prevent="emitInput"
+  >
+    {{ text || value }}
+  </button>
+</template>
```
viniciuskneves, line 13: Another option is to have it as a wrapper so we can avoid spreading styles.

src/components/HdTileSelect.vue
```diff
@@ -0,0 +1,114 @@
+<template>
+  <section class="field tile-select">
```
LarsFe, line 2: Is `field` needed? Right now we have two block elements on one tag.

src/components/HdTileSelect.vue
```diff
@@ -0,0 +1,114 @@
...
+    HdTileSelectItem,
+    HdTileSelectEditableItem,
+  },
+  data() {
+    return {
+      customInputModel: '',
+    };
+  },
+  props: {
+    /*
```
LarsFe, line 43: If it's not used, I think we can remove the whole commented section.

src/components/HdTileSelectEditableItem.vue
```diff
@@ -0,0 +1,97 @@
...
+    :class="{
+      'tile-select-editable-item--selected': isSelected,
+    }"
+    v-bind="$attrs"
+    :value="value"
+    @input="enableEditing"
+  />
+  <input
+    v-else
+    class="tile-select-editable-item tile-select-editable-item__input"
```
LarsFe, line 14: We have a block and an element on the same tag. I think there needs to be made a small adjustment.

src/components/HdTileSelectItem.vue
```diff
@@ -0,0 +1,79 @@
+<template>
+  <button
+    class="tile-select-item"
+    type="button"
+    :class="{
+      'tile-select-item--selected': isSelected,
+    }"
+    @click="emitInput"
```
LarsFe, line 8: We should make the wording more general, as this element is used in different situations. `input` could be enough. Same goes for the next line and line 41.

src/components/HdTileSelectItem.vue
```diff
@@ -0,0 +1,79 @@
...
+  border: 0;
+  padding: $inset-m;
+  font-family: inherit;
+  font-size: inherit;
+  font-weight: inherit;
+  color: inherit;
+
+  &:hover {
+    color: $secondary-color;
+    box-shadow: 0 0 5px 1px rgba(0, 0, 0, .4);
```
LarsFe, line 64: Not for now, but we could think about adding a variable to set pure black color. Makes it just easier to read read without think about the rgb value. `box-shadow: 0 0 5px 1px rgba($black, .4);`

tests/unit/components/HdTileSelectEditableItem.spec.js
```diff
@@ -0,0 +1,43 @@
...
+      data: {
+        isEditing: true,
+      },
+    });
+
+    expect(wrapper.html()).toMatchSnapshot();
+    expect(wrapper.get('input')).toBeTruthy();
+  });
+
+  it('emits value on inpu', () => {
```
LarsFe, line 32: Typo :)

tests/unit/components/__snapshots__/HdTileSelect.spec.js.snap
```diff
@@ -0,0 +1,43 @@
...
+
+exports[`HdTileSelect allows custom new value 1`] = `
+<section class="field tile-select">
+  <section class="tile-select__items"><button type="button" class="tile-select-item tile-select__item">
+      1
+    </button><button type="button" class="tile-select-item tile-select__item">
+      2
+    </button><button type="button" class="tile-select-item tile-select__item">
+      3
+    </button> <button type="button" class="tile-select-item tile-select-editable-item tile-select__item">
```
LarsFe, line 11: Seeing the snapshot build now feels like sth. goes wrong, as we have three classes on the tag. I cannot see
any longer how the parentage is defined with 2 blocks and one element. Same for the other lines and HdTileSelectEditableItem.spec.js.snap

LarsFe:
**Off topic:**
Shouldn't there be an emoji in the title? ✨ 😁
`:sparkles: HdTileSelect component`

src/components/HdTileSelect.vue
viniciuskneves: It comes from here: https://github.com/homeday-de/homeday-blocks/blob/develop/src/styles/inputs.scss#L9

src/components/HdTileSelect.vue
viniciuskneves: I didn't get your point here but it is just to document how `items` can be passed to this component (TS feelings 😅)

src/components/HdTileSelectEditableItem.vue
viniciuskneves: Nice catch!

src/components/HdTileSelectItem.vue
viniciuskneves: I didn't get it. You suggestion is to rename `emitInput` to `input`?

src/components/HdTileSelect.vue
volcanioo: I guess it kinda guides about this variable, and it's not a problem from my side, it's already will be deleted by webpack after rendering process.

src/components/HdTileSelect.vue
volcanioo: In my opinion, we should use `<div>` tags instead of `<section>` element.

> Each `<section>` should be identified, typically by including a heading (`<h1>-<h6>` element) as a child of the `<section>` element.

> Do not use the `<section>` element as a generic container; this is what `<div>` is for, especially when the sectioning is only for styling purposes. A rule of thumb is that a section should logically appear in the outline of a document.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

src/components/HdTileSelectItem.vue
volcanioo: It might be confused, if we rename `emitInput` to `input`.

volcanioo:
I think the title should have the ticket's type instead of emoji. 

> [TYPE] TICKET_TITLE
>
> TYPE [B, F, T] - represents branch purpose (bugfix, feature, task) 
> TICKET_TITLE - is the title of the ticket defined in Jira.

https://homeday.atlassian.net/wiki/spaces/FRO/pages/64880673/Code+standard

viniciuskneves:
> **Off topic:**
> Shouldn't there be an emoji in the title? ✨ 😁
> `:sparkles: HdTileSelect component`

Sad, I've to force the :sparkles: in the PR title =[

src/components/HdTileSelect.vue
LarsFe: Ah ok. Haven't read the text and haven't seen, it's an explanation. Shame on me. 😁  Then it's fine. It just looked like commented code that isn't used and can be removed.

src/components/HdTileSelectItem.vue
LarsFe: Darn. I read `emailInput`. And I read it multiple times. That's weird. Also took me a short moment to see it now. Looks like I have emit-email weakness. `emitInput` is absolutely fine!

viniciuskneves: Hahahahha 👓 😅

src/components/HdTileSelect.vue
viniciuskneves: Done here: https://github.com/homeday-de/homeday-blocks/pull/465/commits/b0c0313c2555b0b58b0fb6f1db2a06f58d7692dd

src/components/HdTileSelect.vue
viniciuskneves: I've added it as `JSDoc` here: https://github.com/homeday-de/homeday-blocks/pull/465/commits/5fb70fbbbf72bb6e79ff2a4060b6ed75128cf9f9

tests/unit/components/HdTileSelectEditableItem.spec.js
viniciuskneves: https://github.com/homeday-de/homeday-blocks/pull/465/commits/ea6a9de697c8065a862787c0fc800f28342088cd

tests/unit/components/__snapshots__/HdTileSelect.spec.js.snap
viniciuskneves: We have this issue as we're scoping each component independently so it will have such conflicts. Doing something like I mentioned here (https://github.com/homeday-de/homeday-blocks/pull/465#discussion_r444054555) might help as we will have a wrapper to deal with base styles.

PR #467, opened 2020-06-22
[B] Fix the input label on Firefox

ilyasmez:
Caught by Percy :purple_heart:

The previous implementation was translating to:
```scss
.field--active .field__label,
.field--double .field__label,
.field--filled .field__label,
.field--select .field__label,
.field__input:-webkit-autofill+.field__label {
   font-size:14px;
   line-height:18px;
   top:5px
}
```

and Firefox was weirdly ignoring the whole block because of the prefix `-webkit-`, even if they are separated with a comma 🤷‍♂️ 

Which resulted in this:
image

##### Always IE11, last time Safari and now Firefox... I love diversity, but not in browsers 😆 or at least, not in the used engines.

### DIFF
--- a/src/styles/inputs.scss
+++ b/src/styles/inputs.scss
@@ -126,10 +126,20 @@ $focusedLabelTop: 5px;
     z-index: 9;
     pointer-events: none;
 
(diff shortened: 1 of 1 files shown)


viniciuskneves:
Btw, Coveralls is stuck in another PR as well 🤷‍♂️

ilyasmez:
> Sorry for that 🙈

I suggested the change 🙊

PR #480, opened 2020-07-08
[T] HdGalleryMedia tests

drapisarda:
https://homeday.atlassian.net/browse/FET-47

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "vue-zoomer": "homeday-de/vue-zoomer#0.5.0"
   },
   "devDependencies": {
+    "@babel/helper-compilation-targets": "^7.10.4",
     "@percy/storybook": "3.2.0",
     "@storybook/addon-actions": "5.2.8",
     "@storybook/addon-console": "1.2.1",
--- a/tests/unit/components/gallery/HDGalleryMedia.spec.js
+++ b/tests/unit/components/gallery/HDGalleryMedia.spec.js
@@ -8,17 +8,9 @@ const wrapperBuilder = wrapperFactoryBuilder(HdGalleryMedia, {
   },
 });
 
-const emptyWrapperBuilder = wrapperFactoryBuilder(HdGalleryMedia, {});
-
 describe('HdGalleryMedia', () => {
   it('renders component', () => {
     const wrapper = wrapperBuilder();
-
-    expect(wrapper.html()).toMatchSnapshot();
-  });
-
-  it('renders empty element', () => {
-    const wrapper = emptyWrapperBuilder();
 
     expect(wrapper.html()).toMatchSnapshot();
   });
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
     "vue-zoomer": "homeday-de/vue-zoomer#0.5.0"
   },
   "devDependencies": {
+    "@babel/helper-compilation-targets": "^7.10.4",
```
viniciuskneves, line 48: What is the reason for this package? I didn't get it.

drapisarda: It was having this error https://stackoverflow.com/questions/62246824/error-err-package-path-not-exported-no-exports-main-resolved-in-app-node-m

viniciuskneves: When are you getting it? Which command?

tests/unit/components/gallery/HDGalleryMedia.spec.js
```diff
@@ -8,10 +8,26 @@ const wrapperBuilder = wrapperFactoryBuilder(HdGalleryMedia, {
...
     const wrapper = wrapperBuilder();
 
     expect(wrapper.html()).toMatchSnapshot();
   });
+
+  it('renders empty element', () => {
+    const wrapper = emptyWrapperBuilder();
+
+    expect(wrapper.html()).toMatchSnapshot();
+  });
```
viniciuskneves, line 24: I would challenge it as well, maybe we can tackle it differently.

Does it make sense to have this component without an item? Shall we go in the direction of having it with required props?

drapisarda: Is there to cover 
`default: () => ({}),`

drapisarda: It does not make much sense is just a little more resilient. 

That test is there to cover HdGalleryMedia:38

viniciuskneves: Yep, I get it that. My only question is if it makes sense. If does, I'm fine but I just don't see it being used like that.

The only place where `HdGalleryMedia` is used is inside `HdGallery` which has a handler for [empty scenario](https://github.com/homeday-de/homeday-blocks/blob/develop/src/components/gallery/HdGallery.vue#L9).

Does it make sense to keep it with a default of `{}`?
IMHO I think the property should be required but I don't know if we have an use case for empty state on this component.

What I'm challenging here is not the implementation itself, it the reasoning behind it.

drapisarda: I do agree with you. I'll give it a refactor time

drapisarda:
done in https://github.com/homeday-de/homeday-blocks/pull/523

Issue #414, opened 2020-04-21
Improve accessibility on HdIcons


metal-gogo:
Without `HdIcon` we could use an `svg` asset like this:

```vue
<template>
  <img
    src="@/assets/img/icons/some.svg"
    width="16"
    height="16"
    alt="Some alt text to describe image"
  >
</template>
```

With `HdIcon` we can make it inline the `svg` with the following code:

```vue
<template>
  <hd-icon
    :src="someIcon"
    width="16px"
    height="16px"
  />
</template>

<script>
import { HdIcon } from 'homeday-blocks';
import { someIcon } from 'homeday-blocks/src/assets/small-icons';

export default {
  name: 'componentName',
  components: {
    HdIcon,
  },
  data() {
    someIcon,
  },
};
</script>
```

This is awesome! This gives us a lot of versatility and styling power. And the best part is that we are getting the icons directly from the design system.

The downside of using the component `HdIcon` on this case is that we lose the accessibility on the image. I believe that we should make a11y a first-class citizen on Homeday Blocks.

I suggest we implement accessibility on `HdIcon` adding the props:

- `title`: string, optional, if not provided could work as an empty `alt` attribute.
- `id`: string, optional, if not provided could be autogenerated
- `description`: string, optional

I add a couple of screenshots from this [accessible svgs](https://css-tricks.com/accessible-svgs/) post describing how could accessibility be implemented 😁.

![image](https://user-images.githubusercontent.com/7268262/79814211-4ad8ef80-8343-11ea-978c-9b122dcc0b1a.png)

![image](https://user-images.githubusercontent.com/7268262/79814186-40b6f100-8343-11ea-99da-57390c5113bb.png)

---

## Reference:

- https://css-tricks.com/accessible-svgs/
- https://getbootstrap.com/docs/4.0/getting-started/accessibility/#interactive-components
- https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML

metal-gogo:
According to the library we are using, we could use a `transformSource` function to add the title.

https://github.com/shrpne/vue-inline-svg/issues/3

PR #481, opened 2020-07-13
✨ make HdIcon accessible

majakomel:
Hi!

This PR improves HdIcon accessibility as suggested in https://github.com/homeday-de/homeday-blocks/issues/414
Changes follow the tips on accessible SVGs from the [linked article](https://css-tricks.com/accessible-svgs/) - the title and desc tags are added to the SVG, along with `aria-labelledby` and `role` attributes.

Looking forward to the review 🙂

### DIFF
--- a/src/components/HdIcon.vue
+++ b/src/components/HdIcon.vue
@@ -1,9 +1,7 @@
 <template>
   <inline-svg
     :src="src"
-    :title="title"
     :id="id"
-    :description="description"
     :transform-source="transform"
     v-bind="$attrs"
   />
@@ -69,24 +67,22 @@ export default {
     },
     addAccessibilityTags(svg) {
       const randId = Math.floor(Math.random() * 1000);
-      const attrs = [];
       if (this.title) {
         const titleId = `${this.id || randId}-title`;
         const titleTag = document.createElementNS('http://www.w3.org/2000/svg', 'title');
-        titleTag.innerHTML = this.title;
+        titleTag.textContent = this.title;
         titleTag.id = titleId;
         svg.insertBefore(titleTag, svg.firstChild);
-        attrs.push(titleId);
+        svg.setAttribute('aria-labelledby', titleId);
       }
       if (this.description) {
         const descId = `${this.id || randId}-desc`;
         const descTag = document.createElementNS('http://www.w3.org/2000/svg', 'desc');
         descTag.id = descId;
+        descTag.textContent = this.description;
         svg.appendChild(descTag);
-        attrs.push(descId);
+        svg.setAttribute('aria-describedby', descId);
       }
-      if (attrs.length) svg.setAttribute('aria-labelledby', attrs.join(' '));
-      svg.setAttribute('role', 'img');
     },
     getFillFromClassNames(classNames) {
       if (!classNames || !this.fillFromClass) {
--- a/src/components/HdIcon.vue
+++ b/src/components/HdIcon.vue
@@ -1,6 +1,9 @@
 <template>
   <inline-svg
     :src="src"
+    :title="title"
+    :id="id"
+    :description="description"
     :transform-source="transform"
     v-bind="$attrs"
   />
@@ -21,6 +24,15 @@ export default {
       type: String,
       required: true,
     },
+    title: {
+      type: String,
+    },
+    id: {
+      type: String,
+    },
+    description: {
+      type: String,
+    },
     // to set a fill based on the path's class
     fillFromClass: {
       type: Object,
@@ -34,6 +46,8 @@ export default {
   },
   methods: {
     transform(svg) {
+      this.addAccessibilityTags(svg);
+
       if (!this.fillFromClass && !this.classFromFill) {
         return svg;
       }
@@ -52,6 +66,27 @@ export default {
       });
 
       return svg;
+    },
+    addAccessibilityTags(svg) {
+      const randId = Math.floor(Math.random() * 1000);
+      const attrs = [];
+      if (this.title) {
+        const titleId = `${this.id || randId}-title`;
+        const titleTag = document.createElementNS('http://www.w3.org/2000/svg', 'title');
+        titleTag.innerHTML = this.title;
+        titleTag.id = titleId;
+        svg.insertBefore(titleTag, svg.firstChild);
+        attrs.push(titleId);
+      }
+      if (this.description) {
+        const descId = `${this.id || randId}-desc`;
+        const descTag = document.createElementNS('http://www.w3.org/2000/svg', 'desc');
+        descTag.id = descId;
+        svg.appendChild(descTag);
+        attrs.push(descId);
+      }
+      if (attrs.length) svg.setAttribute('aria-labelledby', attrs.join(' '));
+      svg.setAttribute('role', 'img');
     },
     getFillFromClassNames(classNames) {
       if (!classNames || !this.fillFromClass) {
--- a/src/components/HdIcon.vue
+++ b/src/components/HdIcon.vue
@@ -44,7 +44,9 @@ export default {
   },
   methods: {
     transform(svg) {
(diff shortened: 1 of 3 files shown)


src/components/HdIcon.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <inline-svg
     :src="src"
+    :title="title"
+    :id="id"
+    :description="description"
```
ilyasmez, line 6: I don't think `title` and `description` are valid SVG attributes.

src/components/HdIcon.vue
```diff
@@ -53,6 +67,27 @@ export default {
 
       return svg;
     },
+    addAccessibilityTags(svg) {
+      const randId = Math.floor(Math.random() * 1000);
+      const attrs = [];
+      if (this.title) {
+        const titleId = `${this.id || randId}-title`;
+        const titleTag = document.createElementNS('http://www.w3.org/2000/svg', 'title');
+        titleTag.innerHTML = this.title;
```
ilyasmez, line 76:
```suggestion
        titleTag.textContent = this.title;
```

`innerHTML` can make the projects vulnerable to XSS attacks.

src/components/HdIcon.vue
```diff
@@ -53,6 +67,27 @@ export default {
...
+        titleTag.id = titleId;
+        svg.insertBefore(titleTag, svg.firstChild);
+        attrs.push(titleId);
+      }
+      if (this.description) {
+        const descId = `${this.id || randId}-desc`;
+        const descTag = document.createElementNS('http://www.w3.org/2000/svg', 'desc');
+        descTag.id = descId;
```
ilyasmez, line 84:
```suggestion
        descTag.id = descId;
        descTag.textContent = this.description;
```
I think you forgot to actually add the content to the element.

src/components/HdIcon.vue
```diff
@@ -53,6 +67,27 @@ export default {
...
+        svg.appendChild(descTag);
+        attrs.push(descId);
+      }
+      if (attrs.length) svg.setAttribute('aria-labelledby', attrs.join(' '));
```
ilyasmez, line 88: Here you're using `aria-labelledby` for both, the title and the description, which is (weirdly) recommended in the article, but I think [aria-describedby](https://www.w3.org/TR/wai-aria/#aria-describedby) will be better for the description, we can keep using `aria-labelledby` for the title though.

src/components/HdIcon.vue
```diff
@@ -53,6 +67,27 @@ export default {
...
+      svg.setAttribute('role', 'img');
```
ilyasmez, line 89: I don't think this is necessary, as it's always added (not covered by the conditions) so it will make more sense to add it to the template directly.

But, I'm personally in favor of not adding it at all, and leave this choice to the consumer of the component as we are binding all the passed attributes with `v-bind="$attrs"`, so if they want to have the role "img", they can just pass it as an attribute:
```vue
<HdIcon src="some-icon.svg" role="img" />
```

tests/unit/components/HdIcon.spec.js
```diff
@@ -84,4 +84,21 @@ describe('HdIcon', () => {
...
+        description: 'Some description.',
+        id: 'icon-id',
+      },
+    });
+
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.contains('title')).toBeTruthy();
+    expect(wrapper.contains('desc')).toBeTruthy();
+    expect(wrapper.attributes('aria-labelledby')).toBe('icon-id-title icon-id-desc');
```
ilyasmez, line 102: Because this manipulated the dom, and we have to test a few things (tags have been added, tags have the right content, the attributes are added and they have the right value...) I think a snapshot test would be perfect here.

src/components/HdIcon.vue
```diff
@@ -53,6 +67,27 @@ export default {
...
+        titleTag.innerHTML = this.title;
+        titleTag.id = titleId;
+        svg.insertBefore(titleTag, svg.firstChild);
+        attrs.push(titleId);
+      }
+      if (this.description) {
+        const descId = `${this.id || randId}-desc`;
+        const descTag = document.createElementNS('http://www.w3.org/2000/svg', 'desc');
+        descTag.id = descId;
+        svg.appendChild(descTag);
```
ilyasmez, line 85: You're using `appendChild` here and `insetBefore` for the titleTag, any reason for this?

majakomel: Yes, it's recommended that the title is the first child because of backward compatibility with SVG 1.1 (https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title). It doesn't seem to matter for desc.

src/components/HdIcon.vue
```diff
@@ -53,6 +65,25 @@ export default {
...
+    addAccessibilityTags(svg) {
+      const randId = Math.floor(Math.random() * 1000);
+      if (this.title) {
+        const titleId = `${this.id || randId}-title`;
+        const titleTag = document.createElementNS('http://www.w3.org/2000/svg', 'title');
+        titleTag.textContent = this.title;
+        titleTag.id = titleId;
+        svg.insertBefore(titleTag, svg.firstChild);
+        svg.setAttribute('aria-labelledby', titleId);
+      }
```
viniciuskneves, line 77: I would suggest you to split `title` and `description` into separated functions just as a matter of structuring the code.

tests/unit/components/HdIcon.spec.js
```diff
@@ -84,4 +84,19 @@ describe('HdIcon', () => {
...
+        title: 'This make me accessible.',
+        description: 'Some description.',
+        id: 'icon-id',
+      },
+    });
+
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.html()).toMatchSnapshot();
+  });
```
viniciuskneves, line 101: I would suggest you to add a test to cover the "random id" case as well.

src/components/HdIcon.vue
volcanioo: Maybe you can use `if` like that, this practice may be more readable

```js
if (this.title) return false;
const randId = Math.floor(Math.random() * 1000);
const titleId = `${this.id || randId}-title`;
const titleTag = document.createElementNS('http://www.w3.org/2000/svg', 'title');
titleTag.textContent = this.title;
titleTag.id = titleId;
svg.insertBefore(titleTag, svg.firstChild);
svg.setAttribute('aria-labelledby', titleId);

```

majakomel:
Sorry for the delay and thanks for suggestions @viniciuskneves & @volcanioo, I reshuffled the code a bit, looks better separated into 2 methods :)

tests/unit/components/HdIcon.spec.js
```diff
@@ -99,4 +99,18 @@ describe('HdIcon', () => {
...
+      props: {
+        src: 'fake/icon5.svg',
+        title: 'This make me accessible.',
+        description: 'Some description.',
+      },
+    });
+
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.html()).toMatchSnapshot();
```
viniciuskneves: 👍

tests/unit/components/HdIcon.spec.js
```diff
@@ -111,6 +111,9 @@ describe('HdIcon', () => {
 
     await wrapper.vm.$nextTick();
 
-    expect(wrapper.html()).toMatchSnapshot();
+    const titleTag = wrapper.find('title')
+    const descTag = wrapper.find('desc')
+    expect(titleTag.element.id).toMatch(/\d+-title/)
+    expect(descTag.element.id).toMatch(/\d+-desc/)
```
viniciuskneves: Ohhh, the random is biting here, right? 🙈

majakomel: Yeah, it didn't fail in the first run :D

PR #490, opened 2020-07-29
[F] Redirect link support added to HDGallery component

volcanioo:
Hello Guys,

I developed the customer-app new design of property tiles. We noticed that the images inside of the gallery carousel are not clickable to redirect the property detail page.

The problem: https://prnt.sc/tqh1h4

### Changes
1. I added a prop to the `HDGallery` component to get the target URL of the image and I dispatched this variable to the `HDGalleryMedia` component. 
2. I added a prop to the `HDGalleryMedia` component to create a `<a>` tag with the target URL. **(Please check the note on below)**
3. I set the z-index structure inside of the HDGallery component.
4. I updated some test snaps

**Note** 
`HDGalleryMedia` is not a standalone component. There is no way to dispatch a prop directly to the `HDGalleryMedia` component. The `HDGalleryMedia` component is working inside of the `HDGallery`.


src/components/gallery/HdGalleryMedia.vue
```diff
@@ -25,6 +25,7 @@
           'background-image': `url('${item.thumbnail}')`,
         }"
       />
+      <a v-if="link" class="gallery-media__object__link" :href="link"></a>
```
leandroinacio, line 28: What if navigation is handled with vue router?
I think would be safer to implement an @click=$emit() and then we can handle redirection from a method.
Also, we need a unit test for this scenario.

volcanioo: working on it

src/components/gallery/HdGallery.vue
```diff
@@ -16,6 +16,7 @@
       <div class="gallery__main">
         <HdGalleryMedia
           :item="currentItem"
+          :link="link"
           :aspect-ratio="aspectRatio"
         />
```
viniciuskneves, line 21: You could `v-bind="$attrs"` here and remove `link` and `aspectRatio` from the `HdGallery` props as they are part of `HdGalleryMedia` props.

I'm not super familiar with `HdGallery` so if it might not make totally sense hehehe

volcanioo: 🐒  hehe, thanks. I'll try use `v-bind`.

volcanioo: I tried but it didn't work. I checked with @aym3nb also and it can't dispatch variables from the parent component. Possible reasons;

1. HDGallery component doesn't have a prop as `item` but `HDGalleryMedia` has it. 
2. `item` prop of the HDGallery component has one watcher. I can't watch the parent component's prop in the child component. 

If you have any thoughts about it just let me know.

metal-gogo:
I'm not sure this should be part of the `HdGallery` component.
For me, it would make sense to set the link wrapping the component.
Something like:

```js
<template>
    <router-link to="route">
        <HdGallery :items="galleryItems"/>
    </router-link>
</template>
```

volcanioo:
@metal-gogo 

`HdGallery` component contains carousel-arrows, carousel-pager, and images. It means we can't use the `<router-link>` as a parent node of the `HDCarousel` because; the `HDCarousel` component contains carousel arrows and carousel pager. We want to add a clickable area for **only images**. We can't use the next and prev buttons or the pager of the carousel if we add a clickable area to the `HDCarousel` directly.

Screenshot 2020-07-30 at 10 43 26

volcanioo:
I did some final changes and I added an event listener directly to the `HDGalleryMedia` component in the `HDGallery` component. In that way, we don't need to dispatch any prop to the component 2 times.

viniciuskneves:
> I'm not sure this should be part of the `HdGallery` component.
> For me, it would make sense to set the link wrapping the component.
> Something like:
> 
> ```js
> <template>
>     <router-link to="route">
>         <HdGallery :items="galleryItems"/>
>     </router-link>
> </template>
> ```

I would like to poke here as well. @volcanioo do we really need to implement it in the HdGallery or is it just for this scenario? If so, @metal-gogo solution solves the problem in a "easier" way. If not, I would poke even further: do users even use the pager in this page? If not, removing the pager, having a single image and wrapping it with router link should be enough.

As I said, I'm just challenging here. If it is really necessary to implement, let's go for it. I've to say that I haven't checked the implementation so far, just dropping some comments here.

volcanioo:
@viniciuskneves 

Thanks, man. Please always keep going to give some feedback to me. I truly care about these comments. 

Actually I changed implementation totally. we don't create any element inside of the component anymore and we are using `onCurrentItemClick` event listener (already have it) to trig click action. 

> Also, the pager and carousel-arrows are a design requirement and we can't remove those things.

viniciuskneves:
> @viniciuskneves
> 
> Thanks, man. Please always keep going to give some feedback to me. I truly care about these comments.
> 
> Actually I changed implementation totally. we don't create any element inside of the component anymore and we are using `onCurrentItemClick` event listener (already have it) to trig click action.
> 
> > Also, the pager and carousel-arrows are a design requirement and we can't remove those things.

@volcanioo as the others already approved, you can proceed with it.

Another input from my side:
> > Also, the pager and carousel-arrows are a design requirement and we can't remove those things.

You can always challenge those decisions =] Especially if they are not data-based.

Nice job here and good that you changed the implementation as we already had something similar :D

PR #507, opened 2020-08-12
[F] Add scaffolding tool

kuroski:
## What's it about

During PEW, I've ported a simple component with @ilyasmez, and we rely in creating scaffolding files that are usually common in every component.

Every new component need:
- A vue/functional component file
- A test file
- A storybook file
- Import the component in `main.js`

## What?

My proposal is to add a tool to automate basic file creation in this project.

In this PR I've added [Hygen](https://www.hygen.io/), based in they're website, Hygen is:

> The scalable code generator that saves you time.

Now we can create a new component by running `npm run new component`

workflow

### DIFF
--- a/_templates/new/component/test.ejs.t
+++ b/_templates/new/component/test.ejs.t
@@ -1,0 +1,26 @@
+---
+to: tests/unit/<%= path || 'components' %>/<%= name %>.spec.js
+---
+import { wrapperFactoryBuilder } from 'tests/unit/helpers';
+import <%= name %> from '@/<%= path %>/<%= name %>.vue';
+
+const wrapperBuilder = wrapperFactoryBuilder(<%= name %>);
+
+describe('<%= name %>', () => {
+  const build = (overrideProps = {}) => {
+    const props = {
+      ...overrideProps,
+    };
+
+    const wrapper = wrapperBuilder({ props });
+
+    return {
+      wrapper,
+    };
+  };
+
+  it('should render', () => {
+    const { wrapper } = build();
+    expect(wrapper.html()).toMatchSnapshot();
+  });
+});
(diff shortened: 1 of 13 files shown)


metal-gogo:
Super cool! Loved it 💜

_templates/new/component/test.ejs.t
```diff
@@ -0,0 +1,26 @@
+---
+to: tests/unit/<%= path || 'components' %>/<%= name %>.spec.js
```
viniciuskneves, line 2: Just out of curiosity, won't `path` be `components` initially by default? Never used Hygen and just want to understand the necessity of such a fallback. You use `path` in `main.ejs.t` and `vue.ejs.t` as well but you don't fallback there, that is why I'm asking.

_templates/new/help/index.ejs.t
```diff
@@ -0,0 +1,5 @@
+---
+message: |
+  - hygen new {bold component} --name [NAME] --path [PATH] (default: components) --functional (default: false)
```
viniciuskneves, line 3: I think you removed the `functional` option.

_templates/new/component/test.ejs.t
kuroski: Thanks, you are totally right, you can check the changes here: c818cc8

I've forgot to update this file, because when I first started doing this PR, I was not resolving the default path before.

_templates/new/help/index.ejs.t
kuroski: Thanks >_<, I've forgot to remove this option 5f01c3c

I've first implemented the component with a `functional` option, but later on I've checked and we don't have any functional component yet.

PR #523, opened 2020-08-26
[T] Test Gallery components

kuroski:
This PR closes ticket [FET-46](https://homeday.atlassian.net/secure/RapidBoard.jspa?rapidView=22&projectKey=FET&modal=detail&selectedIssue=FET-46) and PR #480 

In this PR:

- Create tests for Gallery components
  - [x] HdGallery **(Integration/Unit)**
  - [x] HdGalleryMedia **(Unit)**
  - [x] HdGalleryPlaceholder **(Unit)**
  - [x] HdGalleryCarousel **(Unit)**
- Fix lint error in some of those components
- Change some css selectors names to avoid using more than one element nesting
  - Descending two elements in BEM is not recommended (e.g: `block__element__element2`)
  - If we fall in that situation we can:
    - Rename the selector to something like `block__element-element2`
    - If this doesn't make sense, the component structure need to be re-structured

image

### DIFF
--- a/src/components/gallery/HdGalleryCarousel.vue
+++ b/src/components/gallery/HdGalleryCarousel.vue
@@ -1,39 +1,32 @@
 <template>
   <div
+    class="gallery-carousel"
     :class="{
-      'gallery-carousel': true,
-      isUsingMouse,
-      withPagerInside: pagerInside,
+      'gallery-carousel--with-pager-inside': pagerInside,
     }"
-    @keydown="setUsingMouse(false)"
-    @mousedown="setUsingMouse(true)"
   >
     <div
       class="gallery-carousel__wrap"
       tabindex="0"
       @keydown.self="maybeSelectItem"
-      @blur.self="setUsingMouse(false)"
     >
       <flickity
         ref="flickity"
         :options="flickityOptions"
       >
         <div
+          class="gallery-carousel__item"
           v-for="(item, i) in items"
           :key="i"
           :class="{
-            'gallery-carousel__item': true,
-            'isActive': shouldShowActiveState(i),
+            'is-active': shouldShowActiveState(i),
           } "
         >
-          <div
-            :style="sizerStyles"
-            class="gallery-carousel__item__sizer"
-          />
+          <div :style="sizerStyles" />
 
           <!-- the item.thumbnail field is used as default value for the item image -->
           <!-- IE11 uses this value only because do not support the picture element -->
-          <picture class="gallery-carousel__item__picture">
+          <picture class="gallery-carousel__item-picture">
               <source v-for="(source, media) in item.thumbnailPictureSources"
                 :key="media"
                 :media="`(${media})`" :srcset="source"
@@ -100,7 +93,6 @@ export default {
       currentIndex: 0,
       pagerItemsCount: 0,
       valueChanged: false,
-      isUsingMouse: false,
     };
   },
   computed: {
@@ -216,9 +208,6 @@ export default {
 
       this.$refs.flickity.select(targetSlide);
     },
-    setUsingMouse(usingMouse) {
-      this.isUsingMouse = usingMouse;
-    },
   },
 };
 </script>
@@ -229,10 +218,6 @@ export default {
 .gallery-carousel {
   $_root: &;
   position: relative;
-
-  &.isUsingMouse * {
-    outline: 0;
-  }
 
   &__wrap {
     overflow: hidden;
@@ -251,7 +236,7 @@ export default {
       margin-top: $stack-m;
     }
 
-    #{$_root}.withPagerInside & {
+    #{$_root}--with-pager-inside & {
       position: absolute;
       bottom: 0px;
       width: 100%;
@@ -283,7 +268,7 @@ export default {
 
       &:hover,
       &:focus,
-      &.isActive {
+      &.is-active {
         box-shadow: 0 5px 7px -3px rgba(0, 0, 0, 0.5);
         @media (min-width: $break-tablet) {
           box-shadow: 0 6px 11px -3px rgba(0, 0, 0, 0.45);
@@ -302,7 +287,7 @@ export default {
       object-fit: cover;
     }
 
-    &__picture {
+    &-picture {
       position: absolute;
       top: 0;
       right: 0;
--- a/src/components/gallery/HdGalleryCarousel.vue
+++ b/src/components/gallery/HdGalleryCarousel.vue
@@ -26,7 +26,7 @@
 
           <!-- the item.thumbnail field is used as default value for the item image -->
           <!-- IE11 uses this value only because do not support the picture element -->
-          <picture class="gallery-carousel__item-picture">
+          <picture class="gallery-carousel__picture">
               <source v-for="(source, media) in item.thumbnailPictureSources"
                 :key="media"
                 :media="`(${media})`" :srcset="source"
@@ -213,7 +213,7 @@ export default {
 </script>
 
 <style lang="scss">
-@import 'homeday-blocks/src/styles/mixins.scss';
+@import "homeday-blocks/src/styles/mixins.scss";
 
 .gallery-carousel {
   $_root: &;
@@ -280,14 +280,9 @@ export default {
       width: calc(100% / 7);
       margin-right: $inline-m;
     }
-
-    img {
-      width: 100%;
-      height: 100%;
-      object-fit: cover;
     }
 
-    &-picture {
+  &__picture {
       position: absolute;
       top: 0;
       right: 0;
@@ -297,6 +292,11 @@ export default {
       align-items: center;
       justify-content: center;
       z-index: 1;
+
+    img {
+      width: 100%;
+      height: 100%;
+      object-fit: cover;
     }
   }
 }
--- a/tests/unit/components/gallery/HDGallery.spec.js
+++ b/tests/unit/components/gallery/HDGallery.spec.js
@@ -3,13 +3,15 @@ import HdGallery from '@/components/gallery/HdGallery.vue';
 import HdGalleryPlaceholder from '@/components/gallery/HdGalleryPlaceholder.vue';
 import ITEMS from '@/stories/mocks/GALLERY_ITEMS';
 
+const wrapperFactory = wrapperFactoryBuilder(HdGallery, {
+  props: {
+    items: ITEMS,
+  },
+});
+
 describe('HdGallery', () => {
   const build = (props) => {
-    const wrapper = wrapperFactoryBuilder(HdGallery, {
-      props: {
-        items: ITEMS,
-      },
-    })({ props });
+    const wrapper = wrapperFactory({ props });
 
     return {
       wrapper,
--- a/tests/unit/components/gallery/HDGallery.spec.js
+++ b/tests/unit/components/gallery/HDGallery.spec.js
@@ -1,33 +1,112 @@
+import _ from 'lodash';
 import { wrapperFactoryBuilder } from 'tests/unit/helpers';
 import HdGallery from '@/components/gallery/HdGallery.vue';
+import HdGalleryPlaceholder from '@/components/gallery/HdGalleryPlaceholder.vue';
 import ITEMS from '@/stories/mocks/GALLERY_ITEMS';
 
-const IMAGE_SELECTOR = '.gallery__image';
+describe('HdGalleryCarousel', () => {
+  const build = (props = {}) => {
+    const wrapper = wrapperFactoryBuilder(HdGallery, {
+      props: _.assign({
+        items: ITEMS,
+      }, props),
+    })();
 
-const wrapperBuilder = wrapperFactoryBuilder(HdGallery, {
-  props: {
-    items: ITEMS,
-  },
-});
+    return {
+      wrapper,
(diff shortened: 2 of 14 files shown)


src/components/gallery/HdGalleryCarousel.vue
```diff
@@ -1,39 +1,32 @@
...
-          <div
-            :style="sizerStyles"
-            class="gallery-carousel__item__sizer"
-          />
+          <div :style="sizerStyles" />
 
           <!-- the item.thumbnail field is used as default value for the item image -->
           <!-- IE11 uses this value only because do not support the picture element -->
-          <picture class="gallery-carousel__item__picture">
+          <picture class="gallery-carousel__item-picture">
```
volcanioo, line 29: `picture` class is not a modifier in this case. It's a block element. Are you sure about this change? 
http://getbem.com/naming/
I checked the documentation of BEM, they are saying; `You need to use Element type if the element is semantically tied to its block.`

volcanioo: I searched a little bit I think we can use `gallery-carousel__picture` instead of `gallery-carousel__item__picture` to avoid using more than one element nesting.

Source: https://stackoverflow.com/a/22566593

volcanioo (commented):
Great work! I have some notes about the naming of elements and an alternative idea to avoid using more than one element nesting.

src/components/gallery/HdGalleryCarousel.vue
volcanioo: Another reference: https://medium.com/fed-or-dead/battling-bem-5-common-problems-and-how-to-avoid-them-5bbd23dee319

drapisarda: we can go with _.gallery-carousel__picture_ even just just _.galley-carousel picture_

kuroski: Thanks a lot for the suggestion, and I totally agree with you.

Here are the changes 86be6da 
What do you thinkg?

volcanioo: Bom trabalho!

thanks

volcanioo (approved):
ready

tests/unit/components/gallery/HdGalleryPlaceholder.spec.js
```diff
@@ -0,0 +1,31 @@
...
+import { wrapperFactoryBuilder } from 'tests/unit/helpers';
+import HdGalleryPlaceholder from '@/components/gallery/HdGalleryPlaceholder.vue';
+
+describe('HdGalleryPlaceholder', () => {
+  const build = (props = {}) => {
+    const wrapper = wrapperFactoryBuilder(HdGalleryPlaceholder, {
+      props: _.assign({
+        icon: '',
+        text: '',
+      }, props),
```
viniciuskneves, line 11: The idea from the factory is that you initialize it with the defaults and you can overwrite it in your tests so no need to do this `assign` again.

For example: https://github.com/homeday-de/homeday-blocks/blob/develop/tests/unit/components/form/HdInputFormatter.spec.js#L4-L13 and https://github.com/homeday-de/homeday-blocks/blob/develop/tests/unit/components/form/HdInputFormatter.spec.js#L124-L135

viniciuskneves: The same applies to all other tests. Your factory should be able to create a working wrapper at the end without any extra params. In the tests you can overwrite those defaults if needed.

viniciuskneves: You could have the factory outside the builder, like

```javascript
const wrapperFactory = wrapperFactoryBuilder(HdGalleryPlaceholder, {
      props: {
        icon: '',
        text: '',
      });
```

And use it inside your `build` something like that.

```javascript
const build = (...params) => {
    const wrapper = wrapperFactory(...params);
...
```

The idea is to avoid repeating `props` and everything you need to forward to the factory as it has been already defined somewhere else.

viniciuskneves: Or you could also extend the factory to have something like your `getters` (I will call them getters), could work as well but I'm not sure how it would look like exactly (maybe another PR as it is going to another scope already).

viniciuskneves (commented):
I just dropped a thread, let me know what you think!

Nice job tackling it and coming with some other points as well 👋

tests/unit/components/gallery/HdGalleryPlaceholder.spec.js
kuroski: I see, that makes sense.

In the factory function, we are just overriding the values from props: https://github.com/homeday-de/homeday-blocks/blob/364e324ca88133efaf5438a0f84054b3ed7c7ae8/tests/unit/helpers.js#L51 right?

Before I apply the changes, do you think I should refactor the factory to use the `_.assign`?

I'm questioning this because maybe the prop can be an object. 
And if we just concat the values with spread, we may lose the default values.

e.g:

``` javascript
function factoryBuilder(defaultProps) {
    return (overrideProps) => ({ ...defaultProps, ...overrideProps })
}

const userFactory = factoryBuilder({
    name: 'Daniel',
    item: {
        title: 'Hello World',
        children: []
    }
})

console.log(userFactory())
// { name: 'Daniel', item: { title: 'Hello World', children: [] } }

console.log(userFactory({
    item: {
        children: [1, 2, 3]
    }
}))
// { name: 'Daniel', item: { children: [ 1, 2, 3 ] } }
```

We've lost the default values, so we must explicitly concat the default values from the object

``` javascript
const defaultValues = {
    name: 'Daniel',
    item: {
        title: 'Hello World',
        children: []
    }
}
const userFactory = factoryBuilder(defaultValues)

console.log(userFactory({
    item: {
        ...defaultValues.item
        children: [1, 2, 3]
    }
}))
// { name: 'Daniel', item: { title: 'Hello World', children: [1, 2, 3] } }
```

If we use the `_.assign` that operation is done automatically.

``` javascript
function factoryBuilder(defaultProps) {
    return (overrideProps) => _.assign({ ...defaultProps, ...overrideProps })
}

// ......

console.log(userFactory({
    item: {
        children: [1, 2, 3]
    }
}))
// { name: 'Daniel', item: { title: 'Hello World', children: [1, 2, 3] } }
```

Do you think is relevant?

viniciuskneves: > I see, that makes sense.
> 
> Thanks for the comments man 😄 I'll apply your suggestions.
> 
> In the factory function, we are just overriding the values from props:
> 
> https://github.com/homeday-de/homeday-blocks/blob/364e324ca88133efaf5438a0f84054b3ed7c7ae8/tests/unit/helpers.js#L51
> 
> right?
> Before I apply the changes, do you think I should refactor the factory to use the `_.assign`?
> 
> I'm questioning this because maybe the prop can be an object.
> And if we just concat the values with spread, we may lose the default values.

The "object" case is ignored for props, we have the deepmerge approach in realtor-app (https://github.com/homeday-de/realtor-app/blob/develop/tests/unit/utils.js#L67). I think I tried the `assign` from lodash but had issues using Jest mocks, not 100% sure though hehehe

> Do you think is relevant?

Answering it: yes, I think it is relevant hehehe

kuroski: Awesome!! I will refactor that part then, 1 min.

And I think I will just use the `deepmerge` dependency since we had problem with the `_.assign`

I'm used to use `ramda` instead of `lodash`, they have a https://ramdajs.com/docs/#mergeDeepRight function

kuroski: What do you think?

- 407973d
- 267dfb2
- ad59b5e

viniciuskneves: Rambda seems a nice option. Never used it, just heard. Is it lighter than Lodash? I mean, if we import each function separately?

tests/unit/components/gallery/HDGallery.spec.js
```diff
@@ -1,33 +1,111 @@
...
-
 describe('HdGallery', () => {
-  it('renders component', () => {
-    const wrapper = wrapperBuilder();
+  const build = (props) => {
+    const wrapper = wrapperFactoryBuilder(HdGallery, {
+      props: {
+        items: ITEMS,
+      },
+    })({ props });
```
viniciuskneves, line 12:
```suggestion
  const wrapperFactory = wrapperFactoryBuilder(HdGallery, {
      props: {
        items: ITEMS,
      },
    });
  const build = (...params) => {
    const wrapper = wrapperFactory(...params);
```

It could be something like that. You don't need to create the factory every time. You create it once and reuse as needed.

tests/unit/components/gallery/HDGallery.spec.js
```diff
@@ -1,33 +1,111 @@
...
+      caption: () => wrapper.find('.gallery__caption'),
+      nextButton: () => wrapper.find('.gallery__controls-next'),
+      previousButton: () => wrapper.find('.gallery__controls-prev'),
+      selectedPicture: () => wrapper.find('.gallery-media__object-picture'),
+      selectedPictureImg: () => wrapper.find('.gallery-media__object-picture').find('img'),
+      selectedPictureInfo: () => wrapper.find('.gallery__info'),
+      carousel: () => wrapper.findAll('.gallery__carousel'),
+      carouselWrap: () => wrapper.findAll('.gallery-carousel__wrap'),
+      flickity: () => wrapper.find('.gallery-carousel').find({ ref: 'flickity' }),
+      placeholder: () => wrapper.find(HdGalleryPlaceholder),
```
viniciuskneves, line 25: I thought about those being something from the factory as well. Not sure how it would work but something like:

```javascript
const wrapperFactory = wrapperFactoryBuilder(Component, { props, data... }, { getterFunction1, getterFunction2...});
```

or

```javascript
const wrapperFactory = wrapperFactoryBuilder(Component, { props, data, wrapperGetterFunctions... });
```

viniciuskneves (approved):
Those last comments are just some food for thought.

Nice job here! 🚀

tests/unit/components/gallery/HDGallery.spec.js
kuroski: Awesome suggestion, thanks =D

ce8a2c6

What do you think?

tests/unit/components/gallery/HDGallery.spec.js
kuroski: This is a nice idea, if you want we can discuss it a bit more.

My only worry is, by passing the selectors as parameters to the builder, we are supposed to change the value returned by the factory, and this might change all tests.

What do you think?

viniciuskneves: Hummm... It should be an enhancement only otherwise we will be painful hahah. Maybe it could be injected in the `wrapper`?

Nevertheless, please, do not do it in this PR. We can bring this topic to team sync. It has to be an approach that must of us agree at least.

kuroski: I will write it down and we can bring that in the sync then =D

kuroski:
@viniciuskneves Can you just check this last commit?

d454e17

Because me and @volcanioo discovered a bug with this component, and since I'm testing it, we could fix it right away.

tests/unit/components/gallery/HDGalleryMedia.spec.js
```diff
@@ -22,21 +24,45 @@ describe('HdGalleryMedia', () => {
     expect(wrapper.html()).toMatchSnapshot();
   });
 
-  it('hides the thumbnail when the image is loaded', async () => {
-    const { wrapper, img, thumbnail } = build();
+  it('hides the thumbnail when is not provided', async () => {
```
viniciuskneves, line 27:
```suggestion
  it('hides the thumbnail when it is not provided', async () => {
```

tests/unit/components/gallery/HDGalleryMedia.spec.js
```diff
@@ -22,21 +24,45 @@ describe('HdGalleryMedia', () => {
...
+    const itemWithoutThumb = { ...ITEMS[0], thumbnail: null };
+    const { thumbnail } = build({ item: itemWithoutThumb });
+
+    expect(thumbnail().exists()).toBe(false);
+  });
+
+  it('checks the thumbnail when an item change', async () => {
```
viniciuskneves, line 34: Maybe this title? I was a bit confused with the previous one. Feel free to change it to something else as well.

```suggestion
  it('shows thumbnail when item changes', async () => {
```

viniciuskneves (commented):
Just checked it and seems fine. Can't exactly picture the issue but dropped 2 comments, no blockers from my side.

kuroski:
@viniciuskneves Thanks for the suggestions =D

15ffefa

tests/unit/components/gallery/HdGalleryPlaceholder.spec.js
kuroski: Well, Ramda is a bit lighter, but it is a functional utility library.

Lodash has its functional module that already solves a lot of problems.
But IMHO I think Ramda it's nicer.

But they are just utility libraries xD

PR #525, opened 2020-08-31
[T] update Google Maps API to version 3.42

metal-gogo:
- 📌 (GoogleMaps): update to version 3.42

---

I noticed that on MyHomeday a warning was prompting on the console:

Screen Shot 2020-08-31 at 11 53 21

After taking a look at the [version documentation](https://developers.google.com/maps/documentation/javascript/versions) of Google Maps, it seems that the only versions considered to be currently supported are: `3.40`, `3.41`, and `3.42`.

So I bumped it to `3.42`.

### DIFF
--- a/src/services/googleAPI.js
+++ b/src/services/googleAPI.js
@@ -8,8 +8,10 @@ export function getGoogleAPI(key) {
   }
 
   mapsApiPromise = new Promise((resolve) => {
+    // Check Google Maps version documentation: https://developers.google.com/maps/documentation/javascript/versions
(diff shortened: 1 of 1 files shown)


viniciuskneves (approved):
We could use `quarterly` or `3.nn` as version to avoid manual update but it could lead to issues in the future I think.

kuroski (approved):
🚀

ilyasmez (approved):

volcanioo (approved):

PR #532, opened 2020-09-02
[doc] Document scroll service

drapisarda:
https://homeday.atlassian.net/browse/FET-103

### DIFF
--- a/src/services/README.md
+++ b/src/services/README.md
@@ -219,3 +219,49 @@ const amount = 3;
 const randomInts = generateUniqueNumbers(amount, min, max);
 // [2, 3, 5]
 ```
+
+
+## Scrolling
+We provide a collection of fuctions to manage elements position and the page scrolling.
+
+### elOffset
+Given an element, returns its sizing and position from the top left of the document.
+```javascript
+const fistHeadingPositionAndSize = elOffset(document.querySelector('h1')); 
+// {top: 111.5, left: 25, width: 1352, height: 51}
+```
+
+### getDocumentHeight
+It gives the height of the document.
+```javascript
+const documentHeight = getDocumentHeight(); // 4157
+```
+
+### isWindowFullyScrolled
+Returns a boolean indicating if the document is scrolled until the very bottom.
+```javascript
+const theDocumentIsScrolled = isWindowFullyScrolled(); // false
+```
+
+### freezeScrolling/unfreezeScrolling
+Those functions encapsulates `disableBodyScroll` from [*willmcpo's body-scroll-lock*]((https://github.com/willmcpo/body-scroll-lock)) . Them allows to freeze/unfreeze the document scroll, keeping it allowed on the target element passed as argument.
+
+
+### isElVisible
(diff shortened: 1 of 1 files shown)


src/services/README.md
```diff
@@ -219,3 +219,49 @@ const amount = 3;
...
+```
+
+### isWindowFullyScrolled
+Returns a boolean indicating if the document is scrolled until the very bottom.
+```javascript
+const theDocumentIsScrolled = isWindowFullyScrolled(); // false
+```
+
+### freezeScrolling/unfreezeScrolling
+Those functions encapsulates `disableBodyScroll` from [*willmcpo's body-scroll-lock*]((https://github.com/willmcpo/body-scroll-lock)) . Them allows to freeze/unfreeze the document scroll, keeping it allowed on the target element passed as argument.
```
ilyasmez, line 247:
```suggestion
These functions encapsulate `disableBodyScroll` from [*willmcpo's body-scroll-lock*]((https://github.com/willmcpo/body-scroll-lock)) which supports iOS browsers. They allow freezing/unfreezing the document scroll, keeping it allowed on the target element passed as argument.
```

src/services/README.md
```diff
@@ -219,3 +219,49 @@ const amount = 3;
...
+
+`param {Object} el: the DOM element to check`
+`param {Number} [offset=window.innerHeight]: the value to check the bottom of the element. It's default value has been chose to check, as default, if the element is scrolled past the bottom of the window.`
+
+### scrollToEl
+Scrolls the window to the top of a given element. It accepts as an argument an **object** and uses three properties:
+
+`param {Object} el: the DOM element to scroll to`
+`param {Number} [offset=0]: scrolling offset to the element.`
+`param {Boolean} [onlyIfNotVisible=true]: scrolls to the element only if this is not visible. The check of the visibility is made with isElVisible described above`
```
ilyasmez, line 267:
```suggestion
`param {Boolean} [onlyIfNotVisible=true]: scrolls to the element only if it's is not visible in the viewport. The check of the visibility is made with isElVisible described above`
```

src/services/README.md
```diff
@@ -219,3 +219,49 @@ const amount = 3;
 const randomInts = generateUniqueNumbers(amount, min, max);
 // [2, 3, 5]
 ```
+
+
+## Scrolling
+We provide a collection of fuctions to manage elements position and the page scrolling.
```
ilyasmez, line 225:
```suggestion
We provide a collection of functions to manage elements position and the page scrolling.
```

ilyasmez (commented):
🚀 🤓 
I've just fixed some spelling mistakes and added iOS browsers for the scroll freezing definition as I think it's worth mentioning.

ilyasmez:
It will also look better if you add line breaks between the params of `isElVisible`, `isScrolledPastElBottom` and `scrollToEl`.

volcanioo (approved):

src/services/README.md
drapisarda: Gold

PR #541, opened 2020-09-08
[B] The error message is not updated when the validation rules of the input change

majakomel:
Fix for the ticket: https://homeday.atlassian.net/browse/FET-235

### DIFF
--- a/src/components/form/HdInput.vue
+++ b/src/components/form/HdInput.vue
@@ -154,7 +154,10 @@ export default {
       this.currentType = type;
     },
     required() {
-      this.validate();
+      if (this.error) this.validate();
+    },
+    customRules() {
+      if (this.error) this.validate();
     },
   },
   methods: {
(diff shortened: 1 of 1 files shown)


ilyasmez (commented):
That would do the job for the case of `required` only, I think we can also include `customRules`.
We can also check if there is an error message shown before we run the `validate` again, to avoid showing an error message even if the user did never focus that field (didn't make it dirty in general).

ilyasmez (approved):
That should do the job 👍 🚀 
We can think of a more robust solution depending on the need.

kuroski (approved):
Awesome job =D

In the future, if we have more properties that need to be watched and do the same operation maybe it will make sense to refactor like:
- Trying to create a function to be re-used
- Or create a watcher for multiple properties

But the changes are great =D
:shipit:

drapisarda (approved):

PR #542, opened 2020-09-09
[F] Add indeterminate state to HdCheckbox

majakomel:
This PR closes the ticket: https://homeday.atlassian.net/browse/FET-201

I moved the handling of indeterminate state from [customer-app](https://github.com/homeday-de/customer-app/blob/0517ad46b17d4f81a73e651f4323f8ac71ea9347/src/views/panelViews/DocumentsDataRoom.vue#L30) to homeday-blocks.

I didn't add property updates on the checkbox field (as described [here](https://css-tricks.com/indeterminate-checkboxes/)), since it's useful only for default styling purposes, which we don't need.

### DIFF
--- a/src/components/form/HdCheckbox.vue
+++ b/src/components/form/HdCheckbox.vue
@@ -7,6 +7,7 @@
       'isChecked': isChecked,
       'hasError': !!error,
       'isUsingMouse': isUsingMouse,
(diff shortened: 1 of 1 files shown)


volcanioo (approved):

kuroski (approved):
:shipit:

PR #552, opened 2020-09-21
[F] Support icon button in HdButton

ilyasmez:
https://homeday.atlassian.net/browse/DS-115

~(This is still in discussion with the design team)~ Done!

- Adjusted the styles of the button in the case of icon only.
- The detection of the icon button case happens automatically thanks to MutationObserver, and it checks the content (text) of the button to decide. I went with this solution instead of a prop to avoid redundancy (we can know based on what we have, so no need for a flag).
- Wrapped HdIcon in a span to always make sure its space is reserved when the svg is being loaded, to avoid layout shifts.
- To avoid recalculating the padding and defining the border every time, I switched to using a pseudo-element (`::before`) to display the border of the buttons, and adjust just it's color based on the variant and the state.

### DIFF
--- a/src/components/buttons/HdButton.vue
+++ b/src/components/buttons/HdButton.vue
@@ -95,7 +95,7 @@ export default {
   methods: {
     detectIconButton() {
       // It's an icon button if the content element has no text
-      this.isIconButton = !this.$refs.content.innerText;
+      this.isIconButton = !this.$refs.content.textContent.trim();
     },
   },
 };
--- a/src/components/buttons/HdButton.vue
+++ b/src/components/buttons/HdButton.vue
@@ -3,12 +3,22 @@
     v-on="$listeners"
     :class="computedClasses"
   >
-    <HdIcon
+    <span
       v-if="iconSrc"
-      :src="iconSrc"
       class="btn__icon"
-    />
-    <slot />
+    >
+      <HdIcon
+        :src="iconSrc"
+        width="100%"
+        height="100%"
+      />
+    </span>
+    <span
+      ref="content"
+      class="btn__content"
+    >
+      <slot />
+    </span>
   </button>
 </template>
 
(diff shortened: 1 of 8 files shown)


src/components/buttons/HdButton.vue
```diff
@@ -47,20 +57,45 @@ export default {
...
+          [`${baseClass}--dark-background`]: this.isInDarkBackground,
+          [`${baseClass}--icon-button`]: this.isIconButton,
+        },
+      ];
+    },
+  },
+  mounted() {
+    // We set an observer to watch the content of the button
+    // to know when to switch to the icon button mode
+    this.contentObserver = new MutationObserver(() => {
```
kuroski, line 82: I understood that you are doing this to automatically detect if the button has some content to display.

But I have some questions regarding this:
- Have you tried something like `this.$slots.default`? Because it returns `false` if no slot was provided, so you could replace the MutationObserver
  - Or did you go with this approach to be more strict?
- Is not possible to just use CSS to solve this problem, using something like the [`:empty` selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty)

ilyasmez: Thank you for the review @kuroski!
- Yes, I had the `this.$slots.default` in mind, and you can see me using it in the first attempt [here](https://github.com/homeday-de/homeday-blocks/pull/552/commits/a84ed91d949e8b8043b8e6d10be1891260c08893#diff-3101f32ed834ae53815e7bf66063cc3aR52). The problem I faced with it is that $slots is not reactive, so if the content changes inside after the component is mounted, there is no way to detect that, and this is important because I guess some button will change content based on the viewport. I'm observing only the content part of the button, so that shouldn't cause any performance issues.
- The problem with `:empty` is that it doesn't handle whitespace properly, and won't solve the issue with nested elements.

kuroski: Hmm, I see...
Thanks a lot for the explanation =DDD

kuroski (approved):
Awesome job!
:shipit: 

gif

majakomel (approved):
Looks good!

leandroinacio (approved):

Issue #377, opened 2020-03-20
We shouldn't scope homeday blocks styles with vue's keyword `scoped`.


metal-gogo:
I've seen on our codebase that the components in `homeday-blocks` have scoped styles. For instance: https://github.com/homeday-de/homeday-blocks/blob/develop/src/components/HdToast.vue#L71

However, on [Vue styleguide](https://vuejs.org/v2/style-guide/#Component-style-scoping-essential) says: 

> - Component libraries, however, should prefer a class-based strategy instead of using the scoped attribute.
> - This makes overriding internal styles easier, with human-readable class names that don’t have too high specificity, but are still very unlikely to result in a conflict.

Since we are using [BEM](http://getbem.com/), this could namespace our styles, and with this, we will avoid using `v-deep` in our other protects: https://github.com/homeday-de/customer-app/blob/f170f72fc484dbf6e0500611a24b8326c94b1df4/src/components/buyer-feedback/slides/Finance.vue#L57

PR #559, opened 2020-09-28
[F] Adapt the form components to the new design and refactor them

ilyasmez:
https://homeday.atlassian.net/browse/DS-106

The main purpose to allow passing a status or a clickable icon to the side of the input.
While doing this, I felt that it's better to fix it on the other inputs a s well, to have consistent styling for the form fields.
This is just a start towards a more unified form components, and there are still many things that can be improved: an interface that contains all the API that a form component should have (`validate()`, `showHelper()`, `isFilled`...), because there is a lot of repeated code still. It can have a mixin, move the logic to the base components or maybe use composition API :D 

### What has been changed?
- No more `margin-bottom` to show the helper text, it's now part of the components and it's not absolutely positioned, so it can grow and wrap depending on the length of the text. 
-  Use `tansform` and `opacity` for the transitions instead of properties that cause reflow. 
- Create a `FieldBase` where we can have the shared functionality and styles between all the fields.
- Accept a slot for the right side of the input.
- Create a `TextFieldBase` that contains the shared functionality of the text input and textarea (e.g. clear input button).
- Fill the bottom line instead of using a border which was looking weird on Windows.
 - Adjust the styles to match the new state designs.
 - Drop `inputs.scss` file for a more modular styles (each component contains its styles and the shared styles are in the base components, e.g. FieldBase).
 -  `isFilled` instead of `isEmpty` (to keep things positive 🙂 and keep the defaults as `false` for all the flags).

| Case | Before | After |
|------|--------|------|
| Error state | error-before | error-after |
| Filled state (text) | filled-text-before | filled-text-after |
| Paint | paint-before | paint-after |
| Border on Windows | windows-before | windows-after |

### Next steps
- Use FieldBase for HdCheckbox (same way as in HdRadio).
- Move more shared logic to FieldBase (e.g. Validation).
- Write tests for FieldBase (will be nice, as we can remove the duplicate tests from the from components).
- Add stories for fields with validation

#### Kirby
If the styling of the inputs was based on Block's `inputs.scss`, then you'll have to copy the file manually to your projects and used there, or use the Vue components.
#### All projects
Make sure that you were not using any variables from the deprecated `inputs.scss`, if so, please revisit the parts where it's used.


src/components/form/FieldBase.vue
```diff
@@ -0,0 +1,258 @@
...
+      ['field--errored']: hasError,
+      ['field--active']: active,
+      ['field--filled']: filled,
+      ['field--minimized-label']: minimizedLabel,
+      ['field--disabled']: disabled,
+      ['field--grouped']: grouped,
+    }"
+  >
+    <div
+       v-if="icon"
```
kuroski, line 14: 💅 just lint >_<

```suggestion
      v-if="icon"
```

src/components/form/FieldBase.vue
```diff
@@ -0,0 +1,258 @@
...
+        </label>
+        <div class="field__input-right">
+          <slot name="input-right"/>
+        </div>
+        <div
+          v-if="!grouped"
+          class="field__border"
+        />
+      </div>
+      <p
```
kuroski, line 40: Do you think it makes sense to use a convention and add an `id` here and in the `label` based in the `name` of the input? 

This way when we are rendering an input, we can use `aria-labelledby` and `aria-describedby`.

``` html
<FieldBase
 :name="name"
  ...
>
	<input 
    type="text" 
    :name="name" 
    :id="name" 
    :aria-labelledby="`${name}-label`" 
    :aria-describedby="`${name}-helper`" 
  />
</FieldBase>

File: src/components/form/FieldBase.vue [hunk 3]
```diff
@@ -0,0 +1,258 @@
...
+      default: false,
+    },
+    grouped: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  computed: {
+    helperText() {
+      // We default to an uinvisible character to keep the same element height
```
kuroski, line 100: 💅 

```suggestion
      // We default to an invisible character to keep the same element height
```

src/components/form/TextFieldBase.vue
```diff
@@ -0,0 +1,171 @@
...
+        :src="checkIcon"
+        width="24"
+        height="24"
+        class="text-field__status-icon text-field__status-icon--valid"
+        @click.native="$emit('status-click')"
+      />
+      <button
+        v-else-if="isClearButtonVisible"
+        class="text-field__clear-button"
+        @click="$emit('clear-click')"
```
kuroski, line 25: Maybe just `$emit('clear')` would be enough in this case:

`<TextFieldBase @clear="" />` is already good without the `-click` postfix

src/components/form/HdGoogleAutocomplete.vue
```diff
@@ -112,16 +104,8 @@ export default {
...
-    },
-    fieldClasses() {
-      return {
-        'field--active': this.isActive,
-        'field--filled': !this.isEmpty,
-        'field--invalid': this.isValid === false,
-        'field--disabled': this.disabled,
-      };
+    isFilled() {
+      return this.value !== null && this.value !== undefined && this.value !== '';
```
kuroski, line 108: We are using `lodash` in the project, maybe we could use `_.isEmpty` to handle this

src/components/form/HdInput.vue
```diff
@@ -133,17 +115,8 @@ export default {
...
-    fieldClasses() {
-      return {
-        'field--active': this.isActive,
-        'field--filled': !this.isEmpty,
-        'field--invalid': this.isValid === false,
-        'field--disabled': this.disabled,
-        'field--hasIcon': this.icon,
-      };
+    isFilled() {
+      return this.value !== null && this.value !== undefined && this.value !== '';
```
kuroski, line 119: Same thing here `_.isEmpty`

src/components/form/HdSelect.vue
```diff
@@ -96,24 +94,15 @@ export default {
...
-      return {
-        'field--active': this.isActive,
-        'field--filled': !this.isEmpty,
-        'field--valid': false,
-        'field--invalid': this.isValid === false,
-        'field--disabled': this.disabled,
-        'field--hasIcon': this.icon,
-      };
+    isFilled() {
+      return this.value !== null && this.value !== undefined && this.value !== '';
```
kuroski, line 105: Same thing here `_.isEmpty`

src/components/form/HdSplitInput.vue
```diff
@@ -121,18 +100,9 @@ export default {
...
-    },
-    fieldClasses() {
-      return {
-        'field--active': this.isActive,
-        'field--filled': !this.isEmpty,
-        'field--invalid': this.isValid === false,
-        'field--disabled': this.disabled,
-        'field--hasIcon': this.icon,
-      };
+        .some(value => value !== null && value !== undefined && value !== '');
```
kuroski, line 105: Same thing here `_.isEmpty`

src/components/form/HdTextarea.vue
```diff
@@ -114,18 +103,8 @@ export default {
...
-      return {
-        'field--active': this.isActive,
-        'field--filled': !this.isEmpty,
-        'field--invalid': this.isValid === false,
-        'field--no-label': this.label === '',
-        'field--disabled': this.disabled,
-        'field--hasIcon': this.icon,
-      };
+    isFilled() {
+      return this.value !== null && this.value !== undefined && this.value !== '';
```
kuroski, line 107: Same thing here `_.isEmpty`

kuroski (commented):
Great work amigo 🎉 

A few small comments (really).

src/components/form/FieldBase.vue
ilyasmez: The `<label>` element already has a `for="name"` and the input already has `id="name"`, so `aria-labelledby` is not necessary here. And the helper can't be as a description, because it's not consistent and it will be empty at first in most cases.
Or did I get you wrong?

src/components/form/TextFieldBase.vue
ilyasmez: The thing is that the event that happens is the click, not the clearing, because the clearing happens on the parent level.
We are just reporting that the "clear" button has been clicked (the clearing didn't happen yet). We can also think of a better name, but I'm afraid it will still have the word "click" 😕

src/components/form/FieldBase.vue
kuroski: You are totally right =D
Thanks for the answer 😄

src/components/form/HdGoogleAutocomplete.vue
ilyasmez: `_.isEmpty` is for collections in general, so it wouldn't do the job for us here, for example `_.isEmpty(99)` would be `true`, which is not what we would expect.
And improvement would be to use `_.isNil` instead of checking for `null` and `undefined` separately, but we will still need the check for the empty string.
I remember having this discussion with the rest of the team a long time ago, and we said that keeping it as explicit as it is right now is easier to understand.
But we can also create our own helper to check if the form fields are filled.

src/components/form/TextFieldBase.vue
kuroski: Ook, not a problem 😄

src/components/form/HdGoogleAutocomplete.vue
kuroski: Oh, sorry about that, I didn't know about that conversation =/
And I'm used with `rambda` instead of `lodash` and `isEmpty(99)` in Rambda is `false` 😅 

But it is completely fine to keep explicit =D

kuroski (approved):
:shipit:

majakomel (commented):
This looks great, there's just one thing that is a regression: clicking on down arrow in HdSelect does not open up options anymore (clicking on select input works fine). It works different now that the icon comes from another element.

Screenshot 2020-09-29 at 17 55 22

ilyasmez:
> This looks great, there's just one thing that is a regression: clicking on down arrow in HdSelect does not open up options anymore (clicking on select input works fine). 

Nice catch @majakomel
It should be fixed now: https://github.com/homeday-de/homeday-blocks/pull/559/commits/d3bbf766a35a7c8e2f59d85a19a425602324cfcf
Because `field__input-right` is used for positioning only, I set it's pointer-events to none, but I enable it back for the children, and it's up to the latter to decide if they want to catches pointer events or not.

src/components/form/FieldBase.vue
```diff
@@ -0,0 +1,263 @@
...
+      return this.error || this.helper || '&#8205;';
+    },
+    hasError() {
+      return Boolean(this.error);
+    },
+  },
+};
+</script>
+
+<style scoped lang="scss">
```
leandroinacio, line 111: Shouldn't we avoid scoped css here?
As mentioned on issue #377 , we should try to avoid scoped css to make overriding styles easier.
Specially in this case, where we have a base for the input fields.

ilyasmez: I think this is a bigger topic and we have to tackle it at once (and not scope some styles and some not).
The main reason is that we have to think of a more namespaced classnames, such as `hd-field` instead of `field` which is more likely to be used in other projects and might result in styles leakage.

So I'd say we keep it out of this PRs scope.

ilyasmez:
Hi @kuroski, I resolved the conflicts regarding HdRadio, you can see the component here:
https://frontend-pulls.homeday.de/homeday-blocks/559/?path=/story/components-form-hdradio--default
please have a look at it and let me know if something is broken.
(btw, I think the component is missing the "pressed" state, but that's not related to this PR).

kuroski:
> Hi @kuroski, I resolved the conflicts regarding HdRadio, you can see the component here:
> https://frontend-pulls.homeday.de/homeday-blocks/559/?path=/story/components-form-hdradio--default
> please have a look at it and let me know if something is broken.
> (btw, I think the component is missing the "pressed" state, but that's not related to this PR).

Hello @ilyasmez, I can double-check with Magda, I already asked her about that since the design file is a bit odd, everything is with the same color despite we can see that they are different.

But as a first response, she told me that in the entire design file we were using the same `DB110` color.

You can continue with your task, and if something needs to be fixed in Figma/Component we can create a ticket 😄

kuroski:
@ilyasmez I've talked with Magda, the DS document was outdated, now everything should be fine 😄

ilyasmez:
> Another thing that I noticed: input fields look off when they are pre-populated by the browser ↙️
> 
> Screenshot 2020-10-14 at 17 14 43

Nice catch! We've already fixed this issue before, that's the price of rewriting something 😅 I'll fix it.

aym3nb (commented):
Amazing job as usual @ilyasmez  💯 👏 

While QA'ing I noticed that the clear icon on the right is focusable so when someone is filling a long-form they have to tab twice to go to the next field (same behaviour for fields with visibility toggle)
Screen Shot 2020-10-15 at 16 56 00
Screen Shot 2020-10-15 at 16 44 27

Is this the intended behaviour? because usually, those helpers on the right aren't focusable.

ilyasmez:
@aym3nb 
> While QA'ing I noticed that the clear icon on the right is focusable so when someone is filling a long-form they have to tab twice to go to the next field (same behaviour for fields with visibility toggle)
> Screen Shot 2020-10-15 at 16 56 00
> Screen Shot 2020-10-15 at 16 44 27
> 
> Is this the intended behaviour? because usually, those helpers on the right aren't focusable.

Thanks for QAing the components 🙇 
I made them focusable on purpose because they are actionable, if you check the login form by Google, you'll see that the visibility toggle is actually focusable as well. They are buttons, not just helping icons.

I agree, that they might annoy some users when tabbing between fields, but we can later think of not showing the clear-field button when the input is focused or something like that.

aym3nb (approved):

PR #560, opened 2020-09-28
[T] [DS-102] Use slots for nested components in HdTable

majakomel:
[DS-102](https://homeday.atlassian.net/browse/DS-102)

Instead of passing the object with component definition to the table through body props, a named slot template has to be defined inside the parent component - with property key as a name (depending on which column we want to have a different layout for):

```
<hd-table :header="header" :body="body">
  <template #stars="{value}">
    <HdTagsList :items="value"></HdTagsList>
  </template>
</hd-table>
```

### DIFF
--- a/src/components/HdTable.vue
+++ b/src/components/HdTable.vue
@@ -8,7 +8,7 @@
     <tbody>
       <tr v-for="(tr, i) in body" :key="`row-${i}`">
         <td v-for="(value, i) in tr" :key="`cell-${i}`">
-          <slot :name="i" :value="value">{{ value }}</slot>
+          <slot :name="i" :value="value"><span>{{ value }}</span></slot>
         </td>
       </tr>
     </tbody>
--- a/src/components/HdTable.vue
+++ b/src/components/HdTable.vue
@@ -8,8 +8,7 @@
     <tbody>
       <tr v-for="(tr, i) in body" :key="`row-${i}`">
         <td v-for="(value, i) in tr" :key="`cell-${i}`">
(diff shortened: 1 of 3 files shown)


src/components/HdTable.vue
```diff
@@ -8,8 +8,7 @@
     <tbody>
       <tr v-for="(tr, i) in body" :key="`row-${i}`">
         <td v-for="(value, i) in tr" :key="`cell-${i}`">
```
ilyasmez, line 10:
```suggestion
        <td v-for="(value, name) in tr" :key="`cell-${name}`">
```

src/components/HdTable.vue
```diff
@@ -8,8 +8,7 @@
...
-          <component v-if="isComponent(value)" :is="value.component" v-bind="value.props"></component>
-          <span v-else>{{ value }}</span>
+          <slot :name="i" :value="value"><span>{{ value }}</span></slot>
```
ilyasmez, line 11:
```suggestion
          <slot :name="name" :value="value"><span>{{ value }}</span></slot>
```

ilyasmez (commented):
Looks really nice! 👍  And good job with the first time using scoped-slots!

I've suggested some changes to avoid confusion, because `tr` is an object, and the second param in `v-for` is the name/key, and not the index, so `i` doesn't make sense there, and it's also defined on the upper scope. I know that it was like this before, but it's a good time to fix it 🙂

volcanioo (approved):

ilyasmez (approved):

PR #566, opened 2020-09-29
[T] Update HdRadio component with the design system

kuroski:
https://homeday.atlassian.net/browse/DS-108

In this PR:
- Refactor the HdRadio component to match the design system
- Fix CSS order in the component to avoid having complex selectors
  - E.g: `.radio:not(:focus):not(#{$r}--disabled) .radio__circle` was removed just by reordering the `disabled state`  classes
- Removed the `label` prop, it was not being used in any project that we have at Homeday (This was confirmed with Magda)
- Update the `HdRadio` javascript to "match" @ilyasmez [pull request](https://github.com/homeday-de/homeday-blocks/pull/559)
- Updated `HdRadio` stories

You can see a preview of the changes here:
https://blocks-storybook-6.netlify.app/?path=/story/components-form-hdradio--default

### DIFF
--- a/src/components/form/HdRadio.vue
+++ b/src/components/form/HdRadio.vue
@@ -1,57 +1,53 @@
 <template>
+
   <section
-    :class="wrapperClasses"
-    class="radio-wrapper"
     @keydown="setUsingMouse(false)"
     @mousedown="setUsingMouse(true)"
   >
-    <label
-      v-if="label"
-      :id="`${name}--label`"
-      class="radio-wrapper__label"
-      tabindex="-1"
-    >{{ label }}</label>
     <div
-      :id="name"
-      :aria-labelledby="label ? `${name}--label` : null"
-      class="radio-wrapper__items"
       role="radiogroup"
+      class="radio-wrapper"
+      :class="{
+        'radio-wrapper--errored': hasError,
+        'radio-wrapper--disabled': disabled,
+        'radio-wrapper--vertical': vertical,
+        'radio-wrapper--using-mouse': isUsingMouse,
+      }"
     >
-      <!-- eslint-disable vue/valid-v-on -->
       <div
         v-for="(item, i) in items"
-        :key="`radio-${item.value}`"
+        :key="item.value"
         ref="items"
         :aria-checked="item.value === value ? 'true' : 'false'"
         :class="{
-          'isSelected': item.value === value,
+          'radio--selected': item.value === value,
         }"
         :tabindex="getTabindex(item, i)"
         class="radio"
-        role="radio"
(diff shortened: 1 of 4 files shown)


src/components/form/HdRadio.vue
```diff
@@ -1,57 +1,53 @@
...
         v-for="(item, i) in items"
-        :key="`radio-${item.value}`"
+        :key="item.value"
         ref="items"
         :aria-checked="item.value === value ? 'true' : 'false'"
         :class="{
-          'isSelected': item.value === value,
+          'radio--selected': item.value === value,
         }"
         :tabindex="getTabindex(item, i)"
```
volcanioo, line 25: tab-index always returns `-1` for elements.  fyi
Screenshot 2020-09-29 at 10 23 49

src/components/form/HdRadio.vue
```diff
@@ -218,199 +215,140 @@ export default {
...
-    #{$r}.hasError & {
-      color: $error-color;
-    }
 
-    #{$r}:not(#{$r}--disabled):hover &,
-    #{$r}--active:not(#{$r}--disabled) &,
-    #{$r}--active.hasError:not(#{$r}--disabled) & {
-      color: getShade($secondary-color, 110);
-    }
+  transition: outline $time-s ease-in-out;
```
volcanioo, line 282: I think this code unnecessary. I tried to remove it from the inspector and nothing changed. Please correct me if I'm wrong.

src/components/form/HdRadio.vue
kuroski: Thanks for the comment 😄 

And hmm I didn't change the [original behavior](https://github.com/homeday-de/homeday-blocks/pull/566/files/86ccd7e4d734fe13780bd0ab7bc4133523af643b#diff-e1c07f8f23d0f5161637f99697817ae0R129) or the component, so I don't know exactly why the component is working this way.

I think the `-1` is being used to enable the user to directly focus on the `selected` option.
But in this case, do you think I should remove that and set all tab indexes to 0?

src/components/form/HdRadio.vue
kuroski: Thanks =DD
4fb1418

This CSS actually works, but the animation is so fast that you can't even notice.

leandroinacio (approved):

src/components/form/HdRadio.vue
volcanioo: We shouldn't change anything if we use `tabindex` to directly focus on the `selected` option. All good!

volcanioo (approved):

PR #569, opened 2020-10-02
[T] Add the build per PR action

ilyasmez:
This will deploy each PR in an isolated environment, which will make our reviews fast and easy, especially the UI heavy ones, as it will save us the time to checkout to the head branch and install the dependencies etc. It will also allow us to get a QA from the design team before deploying to production, which I think will be so beneficial in this "redesigning" phase.

### DIFF
--- a/.github/workflows/deploy-pr-to-s3.yml
+++ b/.github/workflows/deploy-pr-to-s3.yml
@@ -46,4 +46,4 @@ jobs:
         with:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           message: |
-            🚀 This PR has been deployed here: https://frontend-pulls.homeday.dev/${{ env.REPO_NAME }}/${{ env.PR_NUMBER }}
+            🚀 This PR has been deployed here: https://frontend-pulls.homeday.de/${{ env.REPO_NAME }}/${{ env.PR_NUMBER }}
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-pr-to-s3.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: ${{ steps.nvm.outputs.NODE_VERSION }}
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build
+        run: npm run build:storybook -- -o ./dist/${{ env.REPO_NAME }}/${{ env.PR_NUMBER }}
```
ilyasmez, line 29: Here I'm using the output param to pass the path where I want the build.
You can check the [docs](https://storybook.js.org/docs/react/workflows/publish-storybook#build-storybook-as-a-static-web-application) form more details.
I used it because they might change something internally to insure the app work in that sub path, but I honestly doubt that, I think a normal build could have done the job for us, but just to keep it sane, I used what they recommend.

.github/workflows/deploy-pr-to-s3.yml
```diff
@@ -0,0 +1,47 @@
...
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          SOURCE_DIR: 'dist'
+          DEST_DIR: '${{ env.REPO_NAME }}/${{ env.PR_NUMBER }}'
+
+      - name: Add comment
+        uses: marocchino/sticky-pull-request-comment@v1
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          message: |
+            🚀 This PR has been deployed here: https://frontend-pulls.homeday.de/${{ env.REPO_NAME }}/${{ env.PR_NUMBER }}/
```
ilyasmez, line 47: The trailing slash here is necessary because Storybook uses relative paths:
Screen Shot 2020-10-02 at 5 23 36 PM
And they don't have a way to specify the base url (like in the Vue apps) unfortunately :/ ... or I just couldn't find it in the docs.

leandroinacio (approved):

volcanioo (approved):

PR #588, opened 2020-10-16
[F] HdForm

ilyasmez:
https://homeday.atlassian.net/browse/FET-238

The features of the component:
- Allows a full customization of the fields. The story showcases one of the cumbersome examples we faced with HdDynamicFrom (@metal-gogo 😉).
- Works with nested fields, so if you have a field that is nested inside another component, the form would still be able to detect it, thanks to [provide/inject](https://v3.vuejs.org/guide/component-provide-inject.html).
- The addition and removal of the fields happens automatically, so you don't have to specify to which form a field belongs.
- The form is smart enough to handle dynamic addition or removal of fields, in case you have some conditional rendering of fields, and will return the form data of just the rendered fields (ehum property management slide ehem... @viniciuskneves @majakomel 😄)
- Had the scroll-to-the-invalid-field feature out of the box, can be enabled with a prop.
- Returns a useful payload, similar to HdDynamicForm's (`isValid`, `formData`, `invalidFields`) in addition to `hasChanged` which can be useful to know if something has changed in the form to avoid unnecessary operations (@kuroski 😉)
- Allows nested form data, if you don't want to lose the structure of your data, you can pass a path as the field's name, e.g. `name="address.city"` will return `{ address: { city: '' } }` as a formData, check the story for a live example.

As we are still using Vue 2, I couldn't run away from using mixins here :/ but it's not as bad because the field components don't use the mixin's methods or data, everything is used just inside the mixin.

Let me know what do you think, questions or suggestions (if it's not a core-feature-request, we can leave it to another PR :) ).

**TODOs**: 
- [x] Test the component


src/components/form/HdForm.vue
```diff
@@ -0,0 +1,90 @@
...
+        return;
+      }
+
+      this.fields.push(fieldToAdd);
+
+      // If a new field was dynamically added to the form, we add its value to the initial form data
+      if (this.initialFormData[fieldToAdd.name] === undefined) {
+        this.initialFormData[fieldToAdd.name] = _cloneDeep(fieldToAdd.value);
+      }
+    },
```
metal-gogo, line 84: I'm not a huge fan of comments. I tend to see them as code smells. I love [this reference](https://refactoring.guru/smells/comments) in particular, I love this statement:

> The best comment is a good name for a method or class.

And, on the `addField` method, I think that this statement applies:

> If a comment explains a section of code, this section can be turned into a separate method via Extract Method. The name of the new method can be taken from the comment text itself, most likely.

So I suggest the following change:

```suggestion
    addField(fieldToAdd) {
      if (_isNil(fieldToAdd.name)) {
        console.warn('The field must have a `name` property to be added to the HdForm', fieldToAdd);
        return;
      }

      this.fields.push(fieldToAdd);
      
      if (this.isDynamicallyAddedField(fieldToAdd.name)) {
        this.initialFormData[fieldToAdd.name] = _cloneDeep(fieldToAdd.value);
      }
    },
    isDynamicallyAddedField(fieldName) {
      return this.initialFormData[fieldName] === undefined;
    },
```

Of course, I don't mean that all comments should be banned. I love [this article](https://css-tricks.com/the-art-of-comments/) by Sarah Drasner where she writes about different kind of comments, and she wrote:

> Code can describe how, but it cannot explain why.

src/components/form/formFieldMixin.js
```diff
@@ -0,0 +1,27 @@
...
+    addFormField: {
+      default: null,
+    },
+    removeFormField: {
+      default: null,
+    },
+  },
+  // We overwrite the injected methods for the children
+  // to avoid adding the children fields to the form as well
+  // in case the component is composed of other form field, e.g. HdInputFormatter
```
metal-gogo, line 12: I love this comment 💜

metal-gogo (commented):
Great work @ilyasmez !
Are we deprecating the `HdDynamicForm` component?
If so, we should add a `console.warn` on the creation of that component saying that it will be deprecated soon.

src/components/form/HdForm.vue
ilyasmez: I like the approach 👍

ilyasmez:
> Are we deprecating the `HdDynamicForm` component?
> If so, we should add a `console.warn` on the creation of that component saying that it will be deprecated soon.

@metal-gogo 
I don't think it makes sense to deprecate it now, because HdForm is not capable of replacing it fully yet, it would make more sense to deprecate it when we create the component that allows us to render fields based on a config/schema. Which would be the next step.

ilyasmez:
Hello @homeday-de/frontend-engineering, with @kuroski's [tests commit](https://github.com/homeday-de/homeday-blocks/pull/588/commits/52c9f82151c4f8c001b830e3a26f561ad96b0318) (❤️) this PR is fully ready to be reviewed, please have a look at it, so we can move forward as it doesn't introduce any breaking changes.

tests/unit/components/form/HdForm.spec.js
```diff
@@ -0,0 +1,194 @@
+import { wrapperFactoryBuilder } from 'tests/unit/helpers';
```
viniciuskneves, line 1: Hey @ilyasmez and @kuroski, wouldn't it be valuable to test the dynamic input component? I think it is quite a critical path, right?

viniciuskneves (commented):
Hey! I loved it ❤️

Sorry for the late feedback ehehehe

I just dropped one comment regarding the test scenarios but in general it is easy to follow.

I just got a bit confused with injecting with default `null` and then checking if the `add/removeFormField` is a function. Couldn't the inject them as functions on `default`? So we could remove the check later?

tests/unit/components/form/HdForm.spec.js
kuroski: I'll check on that, thanks for the comment man 😆

ilyasmez:
@viniciuskneves 
> I just got a bit confused with injecting with default `null` and then checking if the `add/removeFormField` is a function. Couldn't the inject them as functions on `default`? So we could remove the check later?

That's a very valid point, I went with setting it to `null` as it's a falsy value, so we know that these attributes are not provided, an empty function might cause more confusion from my point of **Vue**, because you'll have a function that does nothing as the expected one.

tests/unit/components/form/HdForm.spec.js
ilyasmez: Hey @viniciuskneves, what do you mean by "dynamic input component"? Do you mean HdDynamicForm? Or you're referring to the form components that use the new mixin?

viniciuskneves: Hey @ilyasmez, I meant this: https://github.com/homeday-de/homeday-blocks/pull/588/files#diff-54b6a0b0dcd5da18fce5ff464cfad258f3699fba12d140ef276968d5865f17ffR84

viniciuskneves:
> @viniciuskneves
> 
> > I just got a bit confused with injecting with default `null` and then checking if the `add/removeFormField` is a function. Couldn't the inject them as functions on `default`? So we could remove the check later?
> 
> That's a very valid point, I went with setting it to `null` as it's a falsy value, so we know that these attributes are not provided, an empty function might cause more confusion from my point of **Vue**, because you'll have a function that does nothing as the expected one.

image

Got it. As I said, I got confused as I didn't know about the `default` property itself (I didn't know it could be that customizable with `from`/`default`). I'm fine if the team gets why it is like that.

tests/unit/components/form/HdForm.spec.js
ilyasmez: Ah, alright, got you, that makes sense.

kuroski: b95d87f021ca58ac69398e5cf6a76091f4116d71

Thanks for waiting, there we go @viniciuskneves

viniciuskneves (approved):
Let's just wait for another review before proceeding.

ilyasmez:
This PR has been lingering around for so long, I'll merge as it is an isolated new feature and doesn't introduce any breaking changes, any improvements can be added in new PRs.

PR #592, opened 2020-10-22
[F] Update HdCheckbox component to match the new design

leandroinacio:
https://homeday.atlassian.net/browse/DS-107

### DIFF
--- a/tests/unit/components/form/__snapshots__/HdTextarea.spec.js.snap
+++ b/tests/unit/components/form/__snapshots__/HdTextarea.spec.js.snap
@@ -4,7 +4,7 @@ exports[`HdTextarea is rendered as expected 1`] = `
 <div class="field text-field">
   <!---->
   <div class="field__body">
-    <div class="field__main"><textarea autocomplete="on" id="test name" name="test name" placeholder="" class="textarea" style="height: 100px;"></textarea> <label for="test name" class="field__label">
+    <div class="field__main"><textarea autocomplete="on" id="test name" name="test name" placeholder="" class="textarea" style="height: 100px;"></textarea> <label for="test name" id="test name-label" class="field__label">
         test label
       </label>
       <div class="field__input-right"></div>
--- a/tests/unit/components/form/__snapshots__/HdTextarea.spec.js.snap
+++ b/tests/unit/components/form/__snapshots__/HdTextarea.spec.js.snap
@@ -4,7 +4,7 @@ exports[`HdTextarea is rendered as expected 1`] = `
 <div class="field text-field">
   <!---->
   <div class="field__body">
-    <div class="field__main"><textarea autocomplete="on" id="test name" name="test name" placeholder="" class="textarea" style="height: 100px;"></textarea> <label for="test name" id="test name-label" class="field__label">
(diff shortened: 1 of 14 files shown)


volcanioo (approved):

tests/unit/components/form/__snapshots__/HdTextarea.spec.js.snap
```diff
@@ -4,7 +4,7 @@ exports[`HdTextarea is rendered as expected 1`] = `
 <div class="field text-field">
   <!---->
   <div class="field__body">
-    <div class="field__main"><textarea autocomplete="on" id="test name" name="test name" placeholder="" class="textarea" style="height: 100px;"></textarea> <label for="test name" class="field__label">
+    <div class="field__main"><textarea autocomplete="on" id="test name" name="test name" placeholder="" class="textarea" style="height: 100px;"></textarea> <label for="test name" id="test name-label" class="field__label">
```
viniciuskneves, line 7: Why do we need to add `id` to `label`?
Right now it causes issues with `Textarea` for example as 2 elements will have the same `id` because we don't normalize the `name` when it is used in the `id`. So if we pass a `name` like `test name` we will end up with duplicate ids.
If necessary, to have the `id`, I would recommend normalizing the name first.

viniciuskneves (commented):
The focus state doesn't really reflect the design as it expects to have the same effect as the hover one. The solution are: keep just the outline; keep just the hover state; keep the outline and hover state. I think you can align with design just to keep it consistent with other components as radios have exactly the same issue.

viniciuskneves (commented):
I'm not sure about the indeterminate state because once it is indeterminate, I click it and it goes to selected state but once I click it again it goes back to indeterminate. Shouldn't it go back to unselected? I'm thinking about a form where it is used, usually when they have such a use case the state flows from indeterminate -> selected -> unselected.

ilyasmez:
From what I saw so far, looks really nice @leandroinacio  👍 , I'll try to review the PR today.
Btw, if you open the PR from the main repo (not your fork), you'll be able to trigger the `Deploy the pull request to AWS S3` action, which will provide you with a link that you can already share with the design team so they take care of the QA before going to production.

tests/unit/components/form/__snapshots__/HdTextarea.spec.js.snap
leandroinacio: Not necessary, but nice to have for the [aria-labelby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute).

Can we meet to talk about the duplicated ids? Because I couldn't understand the scenario.

leandroinacio: The focus state was fixed. Should reflect the design now.

leandroinacio: And about the indeterminate state, I'm not sure of the expected behaviour either. I have changed to reflect the scenario you described.

leandroinacio:
@ilyasmez thanks for that info! I will create a new pull request and close this one with a comment tomorrow.

leandroinacio:
New pull request: https://github.com/homeday-de/homeday-blocks/pull/599

PR #599, opened 2020-10-28
[F] Update HdCheckbox component to match the new design

leandroinacio:
https://homeday.atlassian.net/browse/DS-107

### DIFF
--- a/src/components/form/FieldBase.vue
+++ b/src/components/form/FieldBase.vue
@@ -22,10 +22,11 @@
     </div>
     <div class="field__body">
       <div class="field__main">
-        <slot v-bind="{ hasError }" />
+        <slot v-bind="{ hasError, labelId }" />
         <label
           v-if="label"
           :for="name"
+          :id="labelId"
           class="field__label"
         >
           {{ label }}
@@ -48,6 +49,7 @@
 
 <script>
 import HdIcon from 'homeday-blocks/src/components/HdIcon.vue';
+import kebabCase from 'lodash/kebabCase';
 
 export default {
   name: 'FieldBase',
@@ -107,6 +109,12 @@ export default {
     },
     unlabelled() {
       return this.label === '';
+    },
+    labelId() {
+      if (this.name && this.label) {
+        return `${kebabCase(this.name.trim())}-label`;
+      }
+      return null;
     },
   },
 };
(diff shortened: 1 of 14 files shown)


src/components/form/FieldBase.vue
```diff
@@ -108,6 +110,12 @@ export default {
     unlabelled() {
       return this.label === '';
     },
+    labelId() {
+      if (this.name && this.label) {
+        return `${kebabCase(this.name.trim())}-label`;
+      }
+      return null;
+    },
```
metal-gogo, line 118: I don't fully understand this computed property.

If we pass both the `name` and the `label`, then we trim the `name` and discard the `label`, right?

I think this could be simplified to:

```suggestion
    labelId() {
      return `${kebabCase(this.name.trim())}-label`;
    },
```

On one side, the label is only shown if the `label` prop is passed to the component.
We can consider that there will always be a `name` because it is marked as a required prop.

That's why I don't understand why we need to validate the presence of the `label` prop.

src/components/form/FieldBase.vue
```diff
@@ -108,6 +110,12 @@ export default {
     unlabelled() {
       return this.label === '';
     },
+    labelId() {
```
viniciuskneves, line 113: This is the change to tackle the `id` issue I mentioned in the other PR, right?

Ohhh, I just saw the `aria-labelledby` in the component. Now I got it hehehe
As it is something that caused some misunderstandings already, I would recommend to cover it with a test (HdCheckbox case).

src/components/form/FieldBase.vue
viniciuskneves: I think the idea is to avoid having the `labelId` in the `aria-labelledby`, from `HdCheckbox`.

If that is the case, I would add a comment explaining it as it might be weird when revisiting this component in the future. As it is a single case, we don't have `aria-labelledby` in other inputs, it is a bit hard to reason the best structure for that. At the moment if fits our use case.

src/components/form/FieldBase.vue
leandroinacio: Yep, that's it.
Unit test added 👍

volcanioo (approved):
Hey man, great job!

ilyasmez (approved):
Looks good to me 👍 Nice work!
Please feel free to merge once you get the design team's approval.

PR #605, opened 2020-11-05
[RELEASE] Fix HdInput component to not clear on enter

kuroski:
In this release:

- Fix HdInput bug
  - Since we don't specify the clear button type, when the input is in a form if a user hits enter, the clear button is triggered instead of the submit

fbbayg4tmjx0fum22wlm

For more details, check [this link](https://github.com/homeday-de/homeday-blocks/pull/603)


majakomel (approved):
💖

PR #611, opened 2020-11-12
[B][CE-2164] HDInput Active Statement 

volcanioo:
Hey All!

We were discussing the new Input Reset Button feature of homeday-blocks with @metal-gogo and we realized that there is a small missing point regarding input statements on the implementation. This is our meeting’s outcome with @metal-gogo 

* Implementation says the clear button should be shown when the user starts to type something on the input. (https://prnt.sc/vhi0tg).
* The design indicates that the clear button should be shown when there is an error and the input is focused (https://homeday.zeroheight.com/styleguide/s/22414/p/601cda-inputs)

At the moment, we are showing the clear button whenever the field doesn’t show a validation check, and the field is focused. We must show the clear button only when there is an error, and the input is focused. :warning:

### DIFF
--- a/src/components/form/TextFieldBase.vue
+++ b/src/components/form/TextFieldBase.vue
@@ -78,7 +78,7 @@ export default {
   },
   computed: {
     isClearButtonVisible() {
-      return this.$attrs.filled && this.hasFocus;
+      return this.$attrs.filled && this.hasError && this.hasFocus;
     },
     isErrorStatusVisible() {
       return this.hasError && !this.$attrs.active;


metal-gogo (approved):

src/components/form/TextFieldBase.vue
```diff
@@ -78,7 +78,7 @@ export default {
   },
   computed: {
     isClearButtonVisible() {
-      return this.$attrs.filled && this.hasFocus;
+      return this.$attrs.filled && this.hasError && this.hasFocus;
```
viniciuskneves, line 81: As usual, I would recommend adding a test to cover this case 😅

ilyasmez (approved):
We can go with this for now, there is a ticket to research when to show and hide the suffix icons, especially the input cleaning one. 👍

ilyasmez:
@volcanioo If you think this fix is critical for myHomeday. Please merge it a soon as possible, because we are planning to deploy the photo appointment tomorrow, and we depend on some changes in the new Blocks release.
If you don't think you'll manage to find time to write the test that Vini asked for, please just create a followup ticket in the FE board to test the whole component as it doesn't have any tests yet (we can also just add some test cases to `HdInput` and `HdTextarea`, as they are the only components that make use of it).

volcanioo:
Hey @ilyasmez, I see. Probably, this test case will not be finished tomorrow. Let's merge this ticket. I'll create a Tech Dept.

PR #612, opened 2020-11-12
🐛 (Elevations): make sure to create the elevations from 0 to 10

metal-gogo:
Currently while building it prompts the following warning:

Screen Shot 2020-11-12 at 8 07 41

That is because we are trying to use elevations from `0` to `11`.

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -124,7 +124,7 @@ $font-list: ('headline', 'headline'),
 
 
 /** Elevation **/
-@for $level from 0 through length($elevation-levels) - 1 {
+@for $level from 0 to length($elevation-levels) {
   .elevation-#{$level} {
     @include elevation($level);
   }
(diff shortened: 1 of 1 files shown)


kuroski (approved):
Really fast =D

volcanioo (approved):
fast

leandroinacio (approved):

ilyasmez (approved):
Nice catch 👍 

We could also have used `to` instead of `through`, the first is exclusive.

src/styles/main.scss
```diff
@@ -124,7 +124,7 @@ $font-list: ('headline', 'headline'),
 
 
 /** Elevation **/
-@for $level from 0 through length($elevation-levels) {
+@for $level from 0 through length($elevation-levels) - 1 {
```
metal-gogo, line 127:
```suggestion
@for $level from 0 to length($elevation-levels) {
```

PR #620, opened 2020-11-23
👷 Notify slack on releases

viniciuskneves:
It will automatically publish our releases to Slack (`#frontend`).
The message will look like the following:

image

*Secrets already added to the repo*

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,7 +16,6 @@ jobs:
         with:
           node-version: ${{ steps.nvm.outputs.NODE_VERSION }}
           registry-url: https://registry.npmjs.org/
-      - run: npm ci
       - run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_AUTH_TOKEN}}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -50,6 +50,15 @@ jobs:
                     "text": "'"$NOTES"'",
                     "emoji": true
                   }
+                },
+                {
+                  "type": "context",
+                  "elements": [
+                    {
+                      "type": "plain_text",
+                      "text": "Homeday Blocks"
+                    }
+                  ]
                 }
               ]
             }' \
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -9,11 +9,53 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Read .nvmrc
+        run: echo "::set-output name=NODE_VERSION::$(cat .nvmrc)"
+        id: nvm
       - uses: actions/setup-node@v1
         with:
-          node-version: 10.19.0
+          node-version: ${{ steps.nvm.outputs.NODE_VERSION }}
           registry-url: https://registry.npmjs.org/
       - run: npm ci
       - run: npm publish
         env:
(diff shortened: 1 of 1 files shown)


leandroinacio (approved):
This looks really nice 👍

volcanioo (approved):

.github/workflows/publish.yml
```diff
@@ -9,11 +9,53 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Read .nvmrc
+        run: echo "::set-output name=NODE_VERSION::$(cat .nvmrc)"
+        id: nvm
       - uses: actions/setup-node@v1
         with:
-          node-version: 10.19.0
+          node-version: ${{ steps.nvm.outputs.NODE_VERSION }}
```
ilyasmez, line 17: I don't think it's necessary to use the same node version as the app, because we are not building in this action.

And regarding that too (but not related to this PR): I think we can remove the `npm ci` from this action, because `npm publish` doesn't depend on it, and it should make our publishing faster.

.github/workflows/publish.yml
```diff
@@ -9,11 +9,53 @@ jobs:
...
+                    "type": "plain_text",
+                    "text": "'"$NOTES"'",
+                    "emoji": true
+                  }
+                }
+              ]
+            }' \
+            $SLACK_RELEASE_BOT_WEBHOOK_URL
+        env:
+          TITLE: "${{fromJson(steps.get_latest_release.outputs.data).tag_name}} - ${{fromJson(steps.get_latest_release.outputs.data).name}}"
```
ilyasmez, line 59: Do you think it will be helpful to prepend the title with the name of the project to give more context? And to know which project it is in case we add this action to the other projects.

.github/workflows/publish.yml
```diff
@@ -9,11 +9,53 @@ jobs:
...
+          NOTES: "${{fromJson(steps.get_latest_release.outputs.data).body}}"
+          SLACK_RELEASE_BOT_WEBHOOK_URL: ${{ secrets.SLACK_RELEASE_BOT_WEBHOOK_URL }}
```
ilyasmez, line 61: Can you share this secret with us (privately of course) 😁

ilyasmez (approved):
Beautiful! merge merge merge, so we can have a look 😄 

I dropped some tiny little ideas.

.github/workflows/publish.yml
viniciuskneves: How about the note at the bottom like this one
image

.github/workflows/publish.yml
viniciuskneves: Yeah, definitely. Be aware that the secret is just for #frontend channel

.github/workflows/publish.yml
viniciuskneves: Oh, good call 😅

.github/workflows/publish.yml
ilyasmez: Even better! Using the `context` block to "give more context" sounds logical 😄

.github/workflows/publish.yml
ilyasmez: Yes, I assumed so, as there is nothing in the code that defines the channel to use.

viniciuskneves:
Looking forward for the next release, hope it all works fine 😅

PR #630, opened 2020-12-08
[F] HdLink styles

drapisarda:
https://homeday.atlassian.net/browse/DS-111

Yo Kids! I set up the styles right on HdLink but I set the mixins on typography.scss to be used anywhere.
QUESTION: do we want to set the link style as default for the typography i.e. 
```
@mixin hd-typography() {
...
   p a {
      @include link;
   }
...
}
```

?


src/components/HdLink.vue
```diff
@@ -2,6 +2,8 @@
 export const TYPES = [
   'primary',
   'secondary',
+  'dark',
+  'secondary-dark',
 ];
```
ilyasmez, line 7: Instead of having all these cases, I'd suggest to have just `primary` and `secondary`, and set `color: inherit;` in the `link` mixin. That should handle the dark mode without too much hassle.

src/components/HdLink.vue
```diff
@@ -49,3 +51,23 @@ export default {
   },
 };
 </script>
+
+<style lang="scss">
```
ilyasmez, line 55:
```suggestion
<style lang="scss" scoped>
```

We should scope our component styles.
I think you left it unscoped just to allow the us to use `.link` across the app. But to achieve that I recommend adding the styles directly to the `main.scss` to make it more visible and expected.
You can create a separate file, the same way you did for `typography.scss`, maybe call it `link.scss`.

For now you can just import the file in `main.scss` directly (to avoid breaking changes), but later we can make those functional imports optional.

ilyasmez (commented):
The implementation looks good in general, I've left some ideas in the comments.
Please let me know what you think.

viniciuskneves:
Hey!

I really liked it! Looking forward to have it fully used instead of custom links everywhere!

There were some topics discussed in the DS sync meeting but I would like to ask/address another one.

I honestly don't see why we should keep the secondary link. Maybe it has been already discussed with the designers but I don't see when and how it can be used. It is clear, from the image below, that you can't infer that there are multiple links there. 
Can we make it "blue" and "darker blue" when hovered? Or something like that?
image

viniciuskneves:
> > I honestly don't see why we should keep the secondary link. Maybe it has been already discussed with the designers but I don't see when and how it can be used. It is clear, from the image below, that you can't infer that there are multiple links there.
> 
> @viniciuskneves, that's a valid point, which we've also touched in one of the DS syncs. The reason we kept the secondary link, is because it's already used in some places such as our [Homepage](https://www.homeday.de/de/)'s footer or navbar. These links are not meant to be used inside a paragraph, and this is up to the designers to respect it in their design, we don't have to add any rules on the blocks component.
> I'm not sure if you were in that team sync, but we also discussed how we can use the primary link in these cases (especially the navbar) and deprecate the secondary link.

Thanks @ilyasmez. I imagined it had been discussed before but indeed I wasn't present.
I couldn't find examples but the footer is a good one. I was just raising awareness of that. The paragraph case is the one I was worried about but all good!

ilyasmez:
>I honestly don't see why we should keep the secondary link. Maybe it has been already discussed with the designers but I don't see when and how it can be used. It is clear, from the image below, that you can't infer that there are multiple links there.
Can we make it "blue" and "darker blue" when hovered? Or something like that?

@viniciuskneves, that's a valid point, which we've also touched in one of the DS syncs. The reason we kept the secondary link, is because it's already used in some places such as our Homepage's footer or navbar. These links are not meant to be used inside a paragraph, and this is up to the designers to respect it in their design, we don't have to add any rules on the blocks component.
I'm not sure if you were in that team sync, but we also discussed how we can use the primary link in these cases (especially the navbar) and deprecate the secondary link.

(I messed up the comment 🤦  so this comment is just for reference, Vini already replied).

src/styles/main.scss
```diff
@@ -131,3 +131,43 @@ $font-list: ('headline', 'headline'),
     @include elevation($level);
   }
 }
+
+@mixin link {
+  transition: all $time-s ease-in-out;
+  position: relative;
+  // Link underline
+  box-shadow: none; // remove old .link pls
```
ilyasmez, line 139: I don't we do need this anymore

aym3nb (approved):
Nicely done 👍 
Regarding your question, I think it would be great if the links were styled by default.
Also, can you correct a typo in [HdLink docs](https://github.com/homeday-de/homeday-blocks/blob/develop/src/notes/HdLink.md) it says **router-view** instead of **router-link**.

src/styles/main.scss
drapisarda: ok

drapisarda:
> Nicely done 👍
> Regarding your question, I think it would be great if the links were styled by default.
> Also, can you correct a typo in [HdLink docs](https://github.com/homeday-de/homeday-blocks/blob/develop/src/notes/HdLink.md) it says **router-view** instead of **router-link**.

Since you're the only one expressing on this, I'll live the decision to the next meeting.

ilyasmez (approved):
Look noiiice! 👍

PR #647, opened 2021-01-21
[T] Add hasChanged prop to HdForm to make tracking changes easier

majakomel:
I added a new watcher and emitted event to HdForm because we need a way to track if the form has been changed in parent component.

Our specific use case is that we want to confirm with the user if he really wants to leave an unfinished flow via modal, and for that we need to know about the changes.

### DIFF
--- a/src/components/form/HdForm.vue
+++ b/src/components/form/HdForm.vue
@@ -24,10 +24,6 @@ export default {
       type: Boolean,
       default: false,
     },
-    hasChanged: {
-      type: Boolean,
-      default: null,
-    },
   },
   data() {
     return {
@@ -42,14 +38,15 @@ export default {
   },
   created() {
     this.initialFormData = _cloneDeep(this.getFormData());
-
-    if (!_isNil(this.hasChanged)) {
-      this.$watch('fieldsValues', () => {
+  },
+  watch: {
+    fieldsValues() {
+      if (this.$listeners.hasChanged) {
         const formData = formatNestedData(this.getFormData());
         const hasChanged = !_isEqual(formData, this.initialFormData);
-        this.$emit('update:hasChanged', hasChanged);
-      });
-    }
+        this.$emit('hasChanged', hasChanged);
+      }
+    },
   },
   methods: {
     getFormData() {
--- a/src/components/form/HdForm.vue
+++ b/src/components/form/HdForm.vue
@@ -42,7 +42,8 @@ export default {
     if (this.$listeners.hasChanged) {
       this.$watch('fieldsValues', () => {
         const formData = formatNestedData(this.getFormData());
-        const hasChanged = !_isEqual(formData, this.initialFormData);
+        const initialFormData = formatNestedData(this.initialFormData);
+        const hasChanged = !_isEqual(formData, initialFormData);
         this.$emit('hasChanged', hasChanged);
       });
     }
--- a/src/components/form/HdForm.vue
+++ b/src/components/form/HdForm.vue
@@ -24,6 +24,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    hasChanged: {
+      type: Boolean,
+      default: null,
+    },
   },
   data() {
     return {
@@ -31,8 +35,22 @@ export default {
       fields: [],
     };
   },
+  computed: {
+    fieldsValues() {
+      return this.fields.map(field => field.value);
+    },
+  },
   created() {
     this.initialFormData = _cloneDeep(this.getFormData());
+  },
+  watch: {
+    fieldsValues() {
+      if (!_isNil(this.hasChanged)) {
+        const formData = formatNestedData(this.getFormData());
+        const hasChanged = !_isEqual(formData, this.initialFormData);
+        this.$emit('update:hasChanged', hasChanged);
+      }
+    },
   },
   methods: {
     getFormData() {
--- a/src/components/form/HdForm.vue
+++ b/src/components/form/HdForm.vue
@@ -38,15 +38,14 @@ export default {
   },
   created() {
     this.initialFormData = _cloneDeep(this.getFormData());
-
-    if (this.$listeners.hasChanged) {
-      this.$watch('fieldsValues', () => {
-        const formData = formatNestedData(this.getFormData());
-        const initialFormData = formatNestedData(this.initialFormData);
-        const hasChanged = !_isEqual(formData, initialFormData);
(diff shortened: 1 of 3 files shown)


volcanioo (approved):

kuroski (commented):
Maybe two things you could do is:
- Add a story about this property
- Create a `HdForm` test for this new property

😄

src/components/form/HdForm.vue
```diff
@@ -24,15 +24,32 @@ export default {
       type: Boolean,
       default: false,
     },
+    hasChanged: {
+      type: Boolean,
+      default: null,
```
viniciuskneves, line 29: If it is `Boolean`, I would set the default value to `true` or `false`. Do you need `null` somewhere in the code (I didn't find it)?

src/stories/HdForm.stories.js
```diff
@@ -72,6 +74,9 @@ const Template = (args, { argTypes }) => ({
       }));
       console.log('submitted', params);
     },
+    onFormChange(val) {
+      this.hasFormChanged = val;
```
viniciuskneves, line 78: You can also log it in the actions tab, what do you think?

viniciuskneves (commented):
Hey @majakomel I think you need to update the PR description since you went for another solution.

majakomel:
@viniciuskneves You're so fast! I was testing meanwhile if the changes work fine with the feature I'm implementing in myHomeday before pinging for another review. :)

Thanks for initial suggestions and example, I removed the prop as suggested, it was really not necessary.

viniciuskneves (approved):
🚀

src/stories/HdForm.stories.js
```diff
@@ -72,6 +74,10 @@ const Template = (args, { argTypes }) => ({
...
+      console.log('hasChanged', val);
```
kuroski, line 79: Don't forget to remove the `console.logs` 😄

tests/unit/components/form/HdForm.spec.js
```diff
@@ -138,6 +142,34 @@ describe('HdForm', () => {
...
+    await selectors.firstName().setValue('Dave Grohl');
+
+    expect(wrapper.emitted().hasChanged[1][0]).toBe(true);
+
+    await selectors.firstName().setValue(initialFirstNameValue);
+
+    expect(wrapper.emitted().hasChanged[2][0]).toBe(false);
+  });
+
+  test('a HdForm does not emit hasChanged event if there is no listener', async () => {
```
kuroski, line 163: 🚀

src/components/form/HdForm.vue
```diff
@@ -31,8 +31,22 @@ export default {
...
+  computed: {
+    fieldsValues() {
+      return this.fields.map(field => field.value);
+    },
+  },
   created() {
     this.initialFormData = _cloneDeep(this.getFormData());
+
+    if (this.$listeners.hasChanged) {
+      this.$watch('fieldsValues', () => {
```
kuroski, line 43: I don't know if is relevant but maybe we need to make sure to `unwatch` this before the component is destroyed: 

``` javascript
var unwatch = vm.$watch('a', cb)
// later, teardown the watcher
unwatch()
```

I don't know if Vue handles this automatically but maybe it worth testing, otherwise we might face some memory leaks

kuroski (commented):
Just some comments, but for me everything is great 😄 
Nice work!

src/stories/HdForm.stories.js
majakomel: This was actually suggested by Vini, so we can see logs under actions when trying out the component https://github.com/homeday-de/homeday-blocks/pull/647#discussion_r564651065 :)

ilyasmez (commented):
Hi @majakomel,
The implementation will do the just just fine!
But I have a few concerns regarding the design of the component's API.

#### Imperative watcher
I see that you add the watcher only if the listener `hasChanged` is set by the parent **when the component is first created**, I assume that you did so to avoid doing unnecessary computations if the parent is not listening to this event, but that will cause problems if the event listener is dynamically added to the component.
To avoid all this I'd suggest to have a normal declarative watcher, and fire the event whenever a change has been made. The computation time we want to save here is really minor.


#### Name and payload
When I read the event name `hasChange` I'm expecting this event to be fired just once (**the first time** the form data changes), but apparently you're firing this event **every time** something changes, so I'd suggest adapting the name, if you want to keep the same logic you can just rename the event to `change` or `data-change`, so it can be listened to using `@change`, which reads as "on change".
You can return the same payload as we are returning when the form is submitted (you can ignore the complicated `isValid` for now) this will keep the component a little bit clean to avoid adding so many flag-like events, and also because `hasChanged` sounds more like a prop instead of an event (did you look into using the `.sync` modifier instead of having an event?)

majakomel:
@ilyasmez Thanks for suggestions, I changed the watcher to declarative one and renamed the event to `change` to make it more clear.

As for using the `.sync` modifier on the prop, that was my initial PR https://github.com/homeday-de/homeday-blocks/pull/647/commits/f38ef84b9f2fc6ffa438bbf11725a30a9a874dde - but because it still requires an event to be emitted for prop to be updated, @viniciuskneves suggested to remove the use of prop. I honestly don't have any strong preference for either of solutions.

ilyasmez:
> @ilyasmez Thanks for suggestions, I changed the watcher to declarative one and renamed the event to `change` to make it more clear.
> 
> As for using the `.sync` modifier on the prop, that was my initial PR [f38ef84](https://github.com/homeday-de/homeday-blocks/commit/f38ef84b9f2fc6ffa438bbf11725a30a9a874dde) - but because it still requires an event to be emitted for prop to be updated, @viniciuskneves suggested to remove the use of prop. I honestly don't have any strong preference for either of solutions.

Thank you making the changes 🙇 
I would have honestly preferred `.sync` prop if kept the flag-like event, but as we have switched to a generic solution, we won't need it anymore (especially we might end up with many `.sync` props).

I have just one tiny request left, could you please change:
```js
this.$emit('change', hasChanged);
```
To:
```js
this.$emit('change', { hasChanged });
```

The reason is that the second one is more future-proof, so if we want to add any data to the payload later, we won't introduce breaking changes. And it will also make it similar to the on submit payload.

majakomel:
@ilyasmez Done :)

kuroski (approved):

ilyasmez (approved):
🚀 
Nice work!

PR #650, opened 2021-01-22
[B][FET-243] HdTextarea broken by Grammarly

aym3nb:
https://homeday.atlassian.net/browse/FET-243

Currently, when a user clicks in HdTextarea it shrinks and becomes very small becasue of the injected Grammarly element.

As a fix I added `data-gramm_editor="false"` to disable Grammarly for textarea.

ezgif-5-9356df35e6a6

### DIFF
--- a/src/components/form/HdTextarea.vue
+++ b/src/components/form/HdTextarea.vue
@@ -24,6 +24,7 @@
       :style="{ height }"
       :disabled="disabled"
       class="textarea"
+      data-gramm_editor="false"
       @focus="handleFocus"
       @blur="handleBlur"
       @input="handleInput"
--- a/src/components/form/HdTextarea.vue
+++ b/src/components/form/HdTextarea.vue
@@ -29,6 +29,7 @@
       @blur="handleBlur"
       @input="handleInput"
     />
+    <!-- `data-gramm_editor` attribute is used to control Grammarly (chrome extension) -->
   </TextFieldBase>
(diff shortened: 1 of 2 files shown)


volcanioo (approved):

leandroinacio (approved):
Nice catch :)

src/components/form/HdTextarea.vue
```diff
@@ -24,6 +24,7 @@
       :style="{ height }"
       :disabled="disabled"
       class="textarea"
+      data-gramm_editor="false"
```
viniciuskneves, line 27: I would recommend you to add a comment here (or just creating a data where comment would be allowed) explaining this property. If we refactor and loose this commit reference it will be hard to back track.

metal-gogo (approved):

src/components/form/HdTextarea.vue
aym3nb: Good point, I'll add a comment.

src/components/form/HdTextarea.vue
```diff
@@ -29,6 +29,7 @@
       @blur="handleBlur"
       @input="handleInput"
     />
+    <!-- `data-gramm_editor` attribute is used to control Grammarly (chrome extension) -->
```
viniciuskneves, line 32: :+1:

viniciuskneves (approved):

PR #662, opened 2021-02-03
👷  Fix the release workflow and restructure the others

ilyasmez:
There was a problem with the previous implementation, as I wasn't installing the dependencies before running `npm run semantic-release`.

I've also restructured all the workflows (made sure the steps have a descriptive name and added a comment to describe each workflow).

I've adjusted the `announce-on-slack` workflow to use the `release` event payload instead of fetching the latest release manually.

### DIFF
--- a/.github/workflows/announce-release.yml
+++ b/.github/workflows/announce-release.yml
@@ -1,0 +1,53 @@
+# This workflow announces the new release on the Frontend Slack channel
+name: Announce the release on Slack
+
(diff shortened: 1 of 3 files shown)


ilyasmez:
To not bother you with my changes (because I'm in a try-fail-fix phase), I'll merge the changes without waiting for the review.
But I'll still go through the PR creating so you can have a look at the changes later.

PR #676, opened 2021-02-10
✨ New icons

leandroinacio:
https://homeday.atlassian.net/browse/CA-2944

1. Added jsDocs on assets-importer as suggested by @viniciuskneves (just src/assets-importer/index.js)
~~2. Fixed a null pointer on HdCheckbox, when it tries to focus and the element ref is not there (just HdCheckbox.vue)~~
3. New icons added - you can see them under "Medium Icons" (everything else)


src/components/form/HdCheckbox.vue
```diff
@@ -150,7 +150,8 @@ export default {
       this.$emit('input', !this.isChecked);
       this.$nextTick(() => {
         this.validate();
-        this.$refs[this.name].focus();
+        // eslint-disable-next-line no-unused-expressions
+        this.$refs[this.name]?.focus();
```
ilyasmez, line 154: Optional chaining requires a babel plugin, which we're not sure all the projects have.
Also, if this change is not related to the icons, it's better to do it in a separate repo.

src/assets-importer/index.js
```diff
@@ -23,12 +26,21 @@ function fetchFromFigma(endpoint) {
   }).then(res => res.json());
 }
 
+/**
+ * @param  {string} {name - the file name
+ * @param  {RegExp} regex} - the regex used to normalize the file
```
ilyasmez, line 31: Does this format actually work for named params? Did you try to generate the docs and see how does it look like?

I usually use:
```js
/**
 * @param  {String} paramsObject.name - the file name
 * @param  {RegExp} paramsObject.regex - the regex used to normalize the file
 */
```

But I've also read that `$0` gives better output, but I've never tested it:
```js
/**
 * @param  {String} $0.name - the file name
 * @param  {RegExp} $0.regex - the regex used to normalize the file
 */
```

src/assets/small-icons/index.js
```diff
@@ -40,7 +40,6 @@ export { default as eventIcon } from './ic_event.svg';
 export { default as facebookIcon } from './ic_facebook.svg';
 export { default as fanOffIcon } from './ic_fan-off.svg';
 export { default as favoriteBorderIcon } from './ic_favorite-border.svg';
-export { default as favoriteFilledIcon } from './ic_favorite-filled.svg';
```
ilyasmez, line 43: I don't think this is intended.

ilyasmez (commented):
Hey @leandroinacio, I hope you didn't have so much trouble adding the new icons. The process will be even smoother when we move the assets to a separate repo (it's on my TODO list).

And to comply to our new workflow, could you please replace the `[F]` in the title with `✨`?
(If removing the `ic_favorite-filled.svg` was intended then please use `💥`)

src/assets-importer/index.js
leandroinacio: Generated with JSDoc Live Preview on VSCode, looks like this:
image
Using $0 looks exactly like the last one. The params in the method title don't show up, so I think it didn't work.

src/components/form/HdCheckbox.vue
leandroinacio: Aaah bummer 😕

src/assets/small-icons/index.js
leandroinacio: It is not 😅

leandroinacio:
@ilyasmez thanks for the quick response! 
It wasn't complicated, but I spent some time trying to understand why the components I wanted were not being imported.
After @viniciuskneves pointed out I had to update the CONFIG file, then they were still not importing. They were showing up as components in Figma, but Magda had to publish before I actually receive them as components :)

Title updated.

src/components/form/HdCheckbox.vue
leandroinacio: Done, we have a separate PR 👍

src/assets-importer/index.js
ilyasmez: Ah seems like `JSDoc Live Preview` doesn't use the same structuring as the official parser, the official one, nests the params inside, instead of repeating `paramsObject`.

In that case I'd prefer to go for removing the `{` and `}` in the param naming, because it will just cause confusion and it will give a wrong definition in the editor:
Screen Shot 2021-02-10 at 4 45 05 PM
Because the name of the param will be `{name` instead of `name`.

Let's proceed this way until we agree on a way to document named params.

ilyasmez:
> It wasn't complicated, but I spent some time trying to understand why the components I wanted were not being imported.
> After @viniciuskneves pointed out I had to update the CONFIG file, then they were still not importing. They were showing up as components in Figma, but Magda had to publish before I actually receive them as components :)

I see, that you for the info, that would help me build a solution that help us to avoid confusion on the designers' side as well 🙇

ilyasmez (approved):
Thank you @leandroinacio, everything looks great to me.
Please have a look at my comment regarding the jsdocs, other than that, we're good to go.

PR #677, opened 2021-02-10
🐛 Fix undefined pointer on HdCheckbox

leandroinacio:
https://homeday.atlassian.net/browse/CA-2944
Fixed an undefined pointer on HdCheckbox, when it tries to focus and the element ref is not there.

image

### DIFF
--- a/tests/unit/components/form/HdCheckbox.spec.js
+++ b/tests/unit/components/form/HdCheckbox.spec.js
@@ -1,4 +1,4 @@
-import { mount } from '@vue/test-utils';
+import { mount, createLocalVue } from '@vue/test-utils';
 import { wrapperFactoryBuilder } from 'tests/unit/helpers';
 import { getMessages } from '@/lang';
 import HdCheckbox from '@/components/form/HdCheckbox.vue';
@@ -165,7 +165,10 @@ describe('HdCheckbox', () => {
   });
 
   it('At toggle, focus should only trigger if element is available', async () => {
-    jest.spyOn(global.console, 'error');
+    const errorHandler = jest.fn().mockImplementation(() => {});
+    const localVue = createLocalVue({
+      errorHandler,
+    });
 
     const myComponent = {
       components: { HdCheckbox },
@@ -174,12 +177,12 @@ describe('HdCheckbox', () => {
       }),
       template: '<HdCheckbox v-if="isChecked" v-model="isChecked" name="myCheckbox" />',
     };
-    const wrapper = mount(myComponent);
+    const wrapper = mount(myComponent, { localVue });
     const checkbox = wrapper.find(checkboxInnerSelector);
 
     checkbox.trigger('click');
     await wrapper.vm.$nextTick();
 
-    expect(console.error).not.toHaveBeenCalled();
+    expect(errorHandler).not.toHaveBeenCalled();
   });
 });
--- a/tests/unit/components/form/HdCheckbox.spec.js
+++ b/tests/unit/components/form/HdCheckbox.spec.js
@@ -1,3 +1,4 @@
+import { mount } from '@vue/test-utils';
 import { wrapperFactoryBuilder } from 'tests/unit/helpers';
 import { getMessages } from '@/lang';
 import HdCheckbox from '@/components/form/HdCheckbox.vue';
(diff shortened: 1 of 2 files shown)


ilyasmez (approved):
Looks good to me 👍 

I just don't find the test useful, because it tests an error case (because of the `$refs = {}` assignment), instead I would test a case where ref is organically not defined, or remove the test :)

leandroinacio:
@ilyasmez I guess it's good to have it to make sure we catch if someone change that line thinking we will always have the ref to the element, which is not true.
I can add a TODO above it saying to try to reproduce it without forcing the empty object though. Or add a comment describing how the issue happened on sliders. Or both, event better :D 
What do you think?
I just didn't do that yet because I wasn't able to reproduce the empty ref from the unit test code line, but I can try again when I'm done with my other tasks.

tests/unit/components/form/HdCheckbox.spec.js
```diff
@@ -162,4 +162,12 @@ describe('HdCheckbox', () => {
     expect(labelIdAttribute).toContain(computedLabelId);
     expect(ariaLabelledBy).toContain(labelIdAttribute);
   });
+
+  it('At toggle, focus should only trigger if element is available', async () => {
```
kuroski, line 166: I understand this test, but relying on things like `console.log` does not add confidence that this test is actually validating something if we stop `console.logging` things, this test will always pass.

One thing it can be done here is creating something more like the actual user interaction.

The bug happens after the user clicks on the checkbox and then in the next tick the element is focused BUT the element is gone.

A way to reproduce this is by having something like:

``` vue
<template>
  <HdCheckbox
    v-if="isChecked"
    v-model="isChecked"
    name="myCheckbox"
  />
</template>

<script>
export default {
  data: () => ({
    isChecked: true
  })
}
</script>
```

So you could actually build a test that does that, this way it's way clear what you are trying to fix, and it will enforce on why people shouldn't remove this test in the future.

The test could be something like:

``` javascript
it('At toggle, focus should only trigger if element is available', async () => {
    const errorHandler = jest.fn().mockImplementation((err) => {
      console.log(err); // it can be just an empty function
    });

    const localVue = createLocalVue({
      errorHandler,
    });

    const MyComponent = {
      components: { HdCheckbox },
      data: () => ({
        isChecked: true,
      }),
      template: '<HdCheckbox v-if="isChecked" v-model="isChecked" name="myCheckbox" />',
    };

    const wrapper = mount(MyComponent, { localVue });
    const checkbox = wrapper.find(checkboxInnerSelector);

    checkbox.trigger('click');
    await wrapper.vm.$nextTick();

    // in your case, since you fixe it, you can just `expect(errorHandler).not.toHaveBeenCalled()`
    expect(errorHandler).toHaveBeenCalled();
    // this other expect is just here to see that when we don't have your fix, this is the error that logs out
    expect(errorHandler.mock.calls[0][0]).toEqual(new TypeError("Cannot read property 'focus' of undefined"));
})
```

It's a bit more of a boilerplate (I just wrote something that works, it should be nicer to actually refactor the test I'm showing here), but it's a bit clearer what we are doing here.

Explicitly changing component implementation that is **not done** on mount or through "real" user interaction can be a bit more confusing to understand and it might reduce your test confidence in this case.

kuroski (changes requested):
Nice catch!
Please let me know what you think about my suggestion 😄

tests/unit/components/form/HdCheckbox.spec.js
leandroinacio: Sweeeet, that reproduces the issue exactly as expected, thank you very much! 🥳

Just one thing regarding the console.error mock comment I don't understand. You said:
> If we stop console.logging things, this test will always pass or maybe if we revisit this test in the future, it can be hard to find out what is going on.

But we don't console.log anything, the console.error is thrown automatically for null/undefined pointers and the test fails (as the browser would).
This is the error we get with spyOn() (the way it implemented currently):
image

And it passes with the fix:
` PASS  tests/unit/components/form/HdCheckbox.spec.js`

So the spyOn would also fit well here, right? I don't see why not, if I'm missing something please let me know. I don't have any preference between the two implementations, I'm just trying to understand the point you raised. 😅

ilyasmez: The issue with spying on `console.error`, is that it's not reliable, because if there is an [`errorHandler`](https://vuejs.org/v2/api/#errorHandler) that captures the exception, `console.error` won't be called.
Explicitly setting an `errorHandler` like Kuroski did, would be a better solution. You can also just set it to `undefined`, and keep your spy on `console.error`, at least in this case you're sure there is nothing capturing your errors.

leandroinacio: Aaah thanks for clarifying! I still don't agree, but change made. 😄

kuroski (approved):
👍  Nice work

:shipit:

PR #680, opened 2021-02-16
🐛 fix mobile HdDashedList styles

majakomel:
https://homeday.atlassian.net/browse/REAL-3395

The issue was that the values were not shown for items where there was no primary vaue, but secondary value was present. I adjusted the styles to fix the issue.

### Before:
Screenshot_20210125-134720_Samsung Internet

### After:
Screenshot 2021-02-16 at 10 01 10


volcanioo (approved):

leandroinacio (approved):
Looks good to me, and implementation matches the doc on ZeroHeight 👍

PR #708, opened 2021-04-13
✨ Pass row context to the table cell slot

majakomel:
While implementing the [hd-connect dashboard](https://github.com/homeday-de/hd-connect/tree/REAL-3614/dashboard) had the need to access the context of the columns within the cell slot, so we added it.

### DIFF
--- a/src/components/HdTable.vue
+++ b/src/components/HdTable.vue
@@ -8,7 +8,7 @@
     <tbody>
       <tr v-for="(tr, i) in body" :key="`row-${i}`">
         <td v-for="(value, name) in tr" :key="`cell-${name}`">
-          <slot :name="name" :value="value"><span>{{ value }}</span></slot>
+          <slot :name="name" :value="value" :row-context="tr"><span>{{ value }}</span></slot>
         </td>
       </tr>
     </tbody>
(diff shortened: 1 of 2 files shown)


ilyasmez (approved):
Looks good to me!
Just out of curiosity, what's the use case for this?

kuroski:
Heey Ilyas, thanks for reviewing it, so... our use case is the following (This is a minified version):

- We have an array of objects that we are using in the table

``` javascript
const items = [
  {
    uid: string,
    name: string,
    status: string,
    assignee: maybe({
      uid: string,
      name: string
    })
  }
]
```

And in some columns, we have form actions that will trigger a request depending on the user interaction.
So a very naive way to identify that is by having an array of "pending items".

What we do is:
- The user clicks on the table button
- We store the `item uid` in an array of "pending items"
- We update the column item with a loading state it is the `item uid` row

To be more visual, here is an example:

https://user-images.githubusercontent.com/1855125/115028422-1dc30c80-9ec5-11eb-9ad2-7a3c7579442d.mov

So in this case, the column is the `assignees`, which does not contain the `item.uid` property, this is why we did this change.

If you think this change maybe is too much, or it will impact negatively performance-wise, don't worry because I know we can achieve the same behavior by just updating the table item array directly instead of using the approach mentioned above.

ilyasmez:
Thank you for the explanation @kuroski that sounds like a valid use case, another solution would be to return just the index of the item (to match the common practices, especially in array functions), and we leave the work of finding the right item to the developer, but I have nothing against retuning the whole context 👍

PR #715, opened 2021-04-27
💥 Remove the local assets from the project

ilyasmez:
This is a follow up PR to #686, where local assets were deprecated.

If you want to use an asset, you have to use `homeday-assets` instead 🙂

### What does this PR do?

- Deletes the local assets (icons, illustrations and their index files).
- Deletes the `assets-importer` module.
- Removes the `import-assets` npm script.


kuroski (approved):
Best PR 🚀

volcanioo:
EPICCCC!

Screenshot 2021-05-05 at 4 01 21 PM

PR #721, opened 2021-05-03
Unignore dependencies in dependabot's config

ilyasmez:
These rules were introduced by #720, which just translated the rules that we had in the old [Dependabot dashboard](https://app.dependabot.com/).
So I've just unignored all the dependencies, so Dependabot can also open PRs to bump their versions.
I think they were ignored in the past either by accident or laziness :D

This PR intentionally doesn't have an emoji because there is no Gitmoji that matches this case.
🔧  (wrench) can be used but it's meant for project config files and it will create a patch release, which is not needed here.


ilyasmez:
This is a minor change, I'll proceed with merging it, please drop your comments if you have any, I'll follow up on them later.

PR #732, opened 2021-05-06
💥 (hd-modal): updating component to follow DS

metal-gogo:
## Notes

- The actions are no longer an array, but a scope. Following the design, we would have 1 or 2 buttons on the modals. **What do you think about this approach?**
- I removed the `closeIconColor`, `overlayColor`, `width`, `spacingSize`, `noSpacing`, `noExternalSpacing`, and `mobileAlign` props since they doesn't make any sense with the current modal implementation.
- It was not defined in the design reference if a click on the overlay should close the modal, so that's not implemented
- There is still plenty of room to improve the component, specially accessibility-wise, but it was not on the definition of the modal. I see this as potential work for a future iteration:
    - https://dev.to/alenanik/inclusive-components-making-modals-accessible-1hn9
    - https://bitsofco.de/how-and-when-to-use-the-tabindex-attribute/
    - https://bitsofco.de/accessible-modal-dialog/
    - https://www.smashingmagazine.com/2021/03/complete-guide-accessible-front-end-components/#accessible-modals

---

## Tickets:

- https://homeday.atlassian.net/browse/DS-152
- https://www.figma.com/file/kOr0NBglJ1IACXbgKg2QeE/Expos%C3%A9-Preparation-Flow?node-id=4810%3A2

## Design references:

- https://www.figma.com/file/KdznSjw9btqHEKdgmm2UT4co/Design-System-UI-Patterns?node-id=5621%3A6294
- https://www.figma.com/file/kOr0NBglJ1IACXbgKg2QeE/Expos%C3%A9-Preparation-Flow?node-id=4810%3A2


volcanioo:
It looks great for now man. We can talk about other functionalities in the next iteration as we already need to provide this modal ASAP. The only thing that I want to mention is;

* It would be great if we can set `cursor: pointer` styling for the `close button`. The cursor doesn't gives the feeling of it's clickable. 
Screenshot 2021-05-06 at 12 11 29 PM

src/components/HdModal.vue
```diff
@@ -1,139 +1,91 @@
...
-          >
-            {{ action.label }}
-          </button>
-        </div>
+          <slot name="actions">
+            <hd-button
+              modifier="primary"
+              @click="$emit('close')"
+            >
+              OK
```
volcanioo, line 34: Should we update this part?

src/components/HdModal.vue
```diff
@@ -145,151 +97,156 @@ export default {
...
+
+.hd-modal__header {
+  display: flex;
+  // flex-direction: column;
+  // align-items: center;
+  margin-bottom: $sp-xs;
+
+  // @media (min-width: $break-tablet) {
+  //   flex-direction: row;
+  // }
```
volcanioo, line 148: Should we remove comment lines?

src/components/HdModal.vue
```diff
@@ -145,151 +97,156 @@ export default {
...
-  &__body {
-    background-color: white;
-    padding: $sp-l;
+  @media (min-width: $break-tablet) {
+    justify-content: flex-end;
 
-    #{$_root}--spacing-normal & {
-      padding: $sp-m;
+    button {
+      width: auto;
```
volcanioo, line 193: What about `fit-content` instead of `auto` here? Wdyt?

src/components/HdModal.vue
```diff
@@ -145,151 +97,156 @@ export default {
...
 
-  &__header {
-    text-align: center;
-    border-bottom: 1px solid getShade($quaternary-color, 60);
+.hd-modal__close-button {
+  position: absolute;
+  right: $sp-m;
+  top: $sp-m;
+  width: #{$sp-m + $sp-s};
+  height: #{$sp-m + $sp-s};
```
volcanioo, line 227: Nice approach! 
you-rock

volcanioo (commented):
just small comments dude - nice work

ilyasmez:
Very Nice @metal-gogo! The new component looks a loooot better! 👍 

I'm just having doubts about the usage of slots for the actions.

The way the slot is implemented right now will allow us to define how the actions should be displayed, that's cool, but that flexibility comes at a cost, for example, whenever we have more than one action/button —which is a quite common case— we will have to define how they are displayed next to each other (margins, flex...). As this is a common case, I see us having some repeated code within the projects.

As the design specification doesn't show any case where we need a custom footer, what do you think of using scoped slots to allow the developer to customize the way the buttons are displayed while still having the default layouting (margins, flex...) set by the modal component, something similar to what we did to [HdTable](https://github.com/homeday-de/homeday-blocks/blob/master/src/components/HdTable.vue#L10-L12).

I want to also ask about how do you see `isWide` being useful when we have more than one action, but it's related to the slots topic, so let's start with that first.

src/components/HdModal.vue
metal-gogo: It's a default action. Most times it won't be on the component. And "OK" is a common word in German and English.

I don't see a reason to remove it at the moment.

metal-gogo:
> Very Nice @metal-gogo! The new component looks a loooot better! 👍

Thanks ☺️!


> I'm just having doubts about the usage of slots for the actions.

> The way the slot is implemented right now will allow us to define how the actions should be displayed, that's cool, but that flexibility comes at a cost, for example, whenever we have more than one action/button —which is a quite common case— we will have to define how they are displayed next to each other (margins, flex...). As this is a common case, I see us having some repeated code within the projects.
>
> As the design specification doesn't show any case where we need a custom footer, what do you think of using scoped slots to allow the developer to customize the way the buttons are displayed while still having the default layouting (margins, flex...) set by the modal component, something similar to what we did to [HdTable](https://github.com/homeday-de/homeday-blocks/blob/master/src/components/HdTable.vue#L10-L12).
> 
> I want to also ask about how do you see `isWide` being useful when we have more than one action, but it's related to the slots topic, so let's start with that first.

I understand your concerns. I established the base layout styles for the footer following the HdModal definition by the design team on our DS documentation. It already supports having 1 or more items on the footer. 

In this case, I would prefer to have this approach with slots. It provides a ton of flexibility, and at the moment we support the layouts described in our DS out of the box. I made an effort to make it work out of the work with the classes `.hd-modal__footer` and `.hd-modal__footer--is-wide`. 

I'm not closed to refactoring it. Maybe to something similar to what we had on the previous version of the `HdModal` with an array of actions describing the actions and mapping it to an `HdButton`. 

I will be on vacation next week, so @volcanioo can follow-up on this 😀.

volcanioo:
> The way the slot is implemented right now will allow us to define how the actions should be displayed, that's cool, but that flexibility comes at a cost, for example, whenever we have more than one action/button —which is a quite common case— we will have to define how they are displayed next to each other (margins, flex...). As this is a common case, I see us having some repeated code within the projects.

The new implementation will handle the button's positions by the `isWide` attribute, which will decide whether the buttons will be full width or inline. So, in this case; We won't have to set buttons, margins, or flex each time in the default version of the component as the component will do that with the `isWide` attribute. Please let me know what you think about this very specific topic. @ilyasmez 

> what do you think of using scoped slots to allow the developer to customize the way the buttons are displayed while still having the default layouting (margins, flex...) set by the modal component, something similar to what we did to HdTable.

The new implementation is already planned like this. No one will be required to set margins, flex each time of using HDModal in the project. But having slots will provide an opportunity that makes developers available to customize the whole footer without having any element restrictions. If they want to use the checkbox, they can. If they want to use something else, they can. Do you see any risks regarding this in the new implementation? I truly care about your input on that topic! @ilyasmez  

I'm also checking `HdTable`, and I will update the code a bit today. I will let you know in Slack once I'm done.

src/components/HdModal.vue
volcanioo: https://github.com/homeday-de/homeday-blocks/pull/732/commits/14fb11a56c6e06765c1b850f45180b5e03f87a98

ilyasmez:
@volcanioo 
I'm not sure about that 😅  Could you please provide an example where more than one action button is used? I think that will help us discuss this case.... you can add it as a story.

volcanioo:
> @volcanioo
> I'm not sure about that 😅 Could you please provide an example where more than one action button is used? I think that will help us discuss this case.... you can add it as a story.

@ilyasmez https://frontend-pulls.homeday.de/homeday-blocks/732/?path=/story/components-hdmodal--default

ilyasmez:
@volcanioo
> @ilyasmez https://frontend-pulls.homeday.de/homeday-blocks/732/?path=/story/components-hdmodal--default
> Maybe I'm missing something 😅 I don't know - let me know what you think

Very cool! Your recent [commit](https://github.com/homeday-de/homeday-blocks/pull/732/commits/edd141021f6de039fbc5d7ce73970683a438ad20) made it work, it was broken before.

My only concern is that we are imposing some styles on the passed children (`::v-deep`) and that might create styling conflicts.
It might be possible for the user to override the styles for sure, but it's just not transparent enough (the question `where these styles are coming from?` might be asked a lot).

I'd personally prefer to use scoped slots here with a default slot, where we use our layout by default and allow the user to customize the rendering if they want.

But I don't mind going for your solution, as the caveats are quite small. I'd just rename `isButtonsInline` to something generic as `::v-deep > *` will target all the direct children and not just the buttons (e.g. `isFooterInlined`). And let's use the same naming convention for `showCloseIcon` —which might be mistaken for a function—  (e.g. `isClosable` or `isCloseButtonVisible`).


@homeday-de/realtor-expose-squad-frontend @homeday-de/viewings-squad-frontend @homeday-de/property-acquisition-squad-frontend  what do you think?

ilyasmez (approved):
Very nice work @metal-gogo and @volcanioo 👏 👏 👏 🚀 

Please have a look at my last comment regarding the props naming... other than that all good from my side, this is definitely an improvement to what we currently have, so let's go for it now, and we can have other iterations if needed in the future.

volcanioo:
> My only concern is that we are imposing some styles on the passed children (::v-deep) and that might create styling conflicts.

I used the `> *` selector in order to be able to avoid some style conflicts. 

> It might be possible for the user to override the styles for sure, but it's just not transparent enough (the question where these styles are coming from? might be asked a lot).

> I'd personally prefer to use scoped slots here with a default slot, where we use our layout by default and allow the user to customize the rendering if they want.

I implemented that functionality on the train, and I understand your point here. That's fair. I will implement your suggestion to see the pros and cons because I'm also not a fan of `v-deep` if we work on `homeday-blocks`.

> Please have a look at my last comment regarding the props naming... other than that, all good from my side, this is definitely an improvement to what we currently have, so let's go for it now, and we can have other iterations if needed in the future.

Definitely, I will check your comment more carefully today to finish this ticket in a better way! Thanks for your contribution here man! 🚀 

> But I don't mind going for your solution, as the caveats are quite small. I'd just rename isButtonsInline to something generic as ::v-deep > * will target all the direct children and not just the buttons (e.g. isFooterInlined). And let's use the same naming convention for showCloseIcon —which might be mistaken for a function— (e.g. isClosable or isCloseButtonVisible).

Make sense; I will also work on this if we decide to go with this approach. But I want to focus on your suggestion to see the pros and cons before going with the `v-deep` approach.

aym3nb (approved):
Looks great @metal-gogo & @volcanioo  🚀

I'm also okay with going with the current approach then enhancing it further later because it is already so much better than the old one where we had to override its actions every time we used it.

metal-gogo:
Hey @volcanioo , I'm eager to see if it's possible to remove the `v-deep`. Other than that I see everything really nice. And regarding the prop name, I used the `isWide` because on the documentation of the HdModal by the design team they made a distinction of `wide` or `condensed` modals. If that gives a better context to the prop name take it in consideration.

volcanioo:
> Hey @volcanioo , I'm eager to see if it's possible to remove the `v-deep`. Other than that I see everything really nice. And regarding the prop name, I used the `isWide` because on the documentation of the HdModal by the design team they made a distinction of `wide` or `condensed` modals. If that gives a better context to the prop name take it in consideration.

There is no way to handle margins and flex without using `::v-deep` for sure. I only handle margins and display properties by using `>` selector. But I'm working on another approach atm.

volcanioo:
@ilyasmez @aym3nb @metal-gogo 

**Please do not forget to check the console to understand how the functionality works.** I'm very excited to get your comments on the new approach. 

## What I did
* I started to use scoped slots for the actions
* I changed the naming of isButtonsInline

src/components/HdModal.vue
```diff
@@ -1,128 +1,93 @@
...
           >
-            {{ action.label }}
-          </button>
-        </div>
+            <slot :name="name" :value="button">
+              <hd-button
+                :modifier="button.modifier"
+                :is-in-dark-background="button.isInDarkBackground"
+                :disabled="button.disabled"
+                :icon-src="button.iconSrc"
```
ilyasmez, line 39:
```suggestion
                v-bind="button"
```

To give more flexibility here, I think it's better to just pass everything to the button component, so in case we add a new prop to the button we don't have to update this.

src/components/HdModal.vue
```diff
@@ -1,128 +1,93 @@
...
+                @click="$emit('action', button.name)"
```
ilyasmez, line 40:
```suggestion
                @click="$emit('action', button)"
```

Let's just send the whole object as a payload, just to avoid using `actions.find()` in the parent component in case the developer needs some data from the object itself.

src/components/HdModal.vue
```diff
@@ -1,128 +1,93 @@
...
-              'btn--primary': action.type === 'primary',
-              'btn--ghost': action.type !== 'primary',
-            }"
-            class="modal__actions__action btn"
-            type="button"
-            @click="action.callback"
+          <div
+            class="hd-modal__action"
+            v-for="(button, name) in actions"
+            :key="`action-${name}`"
```
metal-gogo, line 32: I saw the styles added by the `hd-modal__action` class, and it mostly adds some styling for layout purposes. Maybe we could move those styles directly to the `<hd-button>` and it could work? I think it's best to avoid adding the `<div>` if possible.

src/components/HdModal.vue
```diff
@@ -1,128 +1,93 @@
...
           >
-            {{ action.label }}
-          </button>
-        </div>
+            <slot :name="name" :value="button">
```
metal-gogo, line 34: Is there any particular reason to keep this `<slot>`? Since we are accepting an array prop with actions, I think we don't longer need the slot. Maybe I'm missing something :).

src/components/HdModal.vue
volcanioo: I have to put `hd-button` inside of the `<slot>` for this reason, I had to use a div by using `hd-modal__action`. Let me know if you have another suggestion

src/components/HdModal.vue
volcanioo: I used to be able to use `scoped slots`

src/components/HdModal.vue
```diff
@@ -1,128 +1,93 @@
...
+              <hd-button
+                :modifier="button.modifier"
+                :is-in-dark-background="button.isInDarkBackground"
```
metal-gogo, line 37: I think we can skip the `is-in-dark-background` prop. The design of the modal always has a light background. At least for the current design.

volcanioo: `is-in-dark-background` one of the attributes of the HDButton, I'd prefer to give this support as well as use `hd-button` with full functionality for future iterations.

src/components/HdModal.vue
```diff
@@ -1,128 +1,93 @@
...
+      default: true,
     },
-    mobileAlign: {
+    lang: {
       type: String,
-      default: 'center',
-      validator: (mobileAlign) => ['center', 'bottom'].includes(mobileAlign),
+      default: 'de',
     },
     actions: {
```
ilyasmez, line 92: For the new solution to work properly, the all the entries of `actions` should have a "name" attribute.
To make this clear for the developer, I suggest to add a custom [prop validator](https://vuejs.org/v2/guide/components-props.html#Prop-Validation) here. Something like: `(entry) => typeof entry.name == 'string';` can do the job, you can also show a warning to explain why it's not valid (e.g. `The attribute "name" is required`)

src/components/HdModal.vue
volcanioo: https://vuejs.org/v2/guide/components-slots.html#Scoped-Slots

ilyasmez (commented):
Super nice @volcanioo, scoped slots will give us more flexibility without forcing any styles for sure!

I've dropped some comments and suggestions, please have a look at them.

Also, to give the developer total control over the footer —as the previous solution tried to achieve— you can even wrap the content of the `<footer>` element inside a `<slot name="footer">` (while keeping what you currently have as default), this way, if the developer wants to fully customize the footer without using the `actions` prop, they will be able to do so.

src/components/HdModal.vue
```diff
@@ -1,128 +1,93 @@
...
-      validator: (spacingSize) => ['large', 'normal'].includes(spacingSize),
+      default: '',
     },
-    noSpacing: {
+    isWide: {
       type: Boolean,
       default: false,
     },
-    noExternalSpacing: {
+    showCloseIcon: {
```
aym3nb, line 84: Also, don't forget Ilyas's previous suggestion about renaming this prop.

> And let's use the same naming convention for showCloseIcon —which might be mistaken for a function— (e.g. isClosable or isCloseButtonVisible).

src/components/HdModal.vue
metal-gogo: What do you think on passing everything that comes on the action to the `hd-button` with a `v-bind`? And just add the @click. This way we don't need to explicitly pass all the props. What do you think?

src/components/HdModal.vue
```diff
@@ -134,6 +99,14 @@ export default {
       closeIcon,
     };
   },
+  computed: {
+    withIcon() {
+      return this.iconSrc.length > 0;
+    },
+    t() {
```
metal-gogo, line 106: I don't see this computed property used in the component. Maybe we could remove it??

src/components/HdModal.vue
volcanioo: On it

src/components/HdModal.vue
volcanioo: https://github.com/homeday-de/homeday-blocks/pull/732/commits/76ac2a99328fe24f43d688b54a190468aa7fa71e

src/components/HdModal.vue
volcanioo: https://github.com/homeday-de/homeday-blocks/pull/732/commits/397b212782decb1480f40b224cab14fc9303f453

src/components/HdModal.vue
volcanioo: https://github.com/homeday-de/homeday-blocks/pull/732/commits/9fa03222c4b4165db9c1b831f9dd2748d4e89a83 @ilyasmez - thanks for reminding me bro <3

src/components/HdModal.vue
volcanioo: https://github.com/homeday-de/homeday-blocks/pull/732/commits/664aea4e6734685f1edf75779d57f89f88861898

volcanioo:
> Also, to give the developer total control over the footer —as the previous solution tried to achieve— you can even wrap the content of the `<footer>` element inside a `<slot name="footer">` (while keeping what you currently have as default), this way, if the developer wants to fully customize the footer without using the actions prop, they will be able to do so.

That was a great suggestion @ilyasmez - thanks for that! Here are the changes: https://github.com/homeday-de/homeday-blocks/pull/732/commits/411c2b8557b273a47e4b7becea4242d8fc817560

* I also added another version of HDModal into stories to show we also support a custom footer.
* We are not adding our stylings to the footer if it's custom.

src/components/HdModal.vue
volcanioo: https://github.com/homeday-de/homeday-blocks/commit/76ac2a99328fe24f43d688b54a190468aa7fa71e

src/components/HdModal.vue
ilyasmez: Very nice @volcanioo,
Just one little improvement to your implementation:
You can make use of the `validator` function even more, by returning `false` if there is an issue with the prop, printing the error is good, but it's even better to use what Vue provides (so it can show the error to the developer in a better way and also fail the build if needed). You can write the validator like this:

```js
validator: (actions) => {
  const indexOfActionMissingName = actions.findIndex((action) => typeof action.name !== 'string');
  
  if (indexOfActionMissingName >= 0) {
    console.warn(`The attribute "name" is required on actions[${indexOfActionMissingName}]`);
    return false;
  }
  
  return true;
},
```

volcanioo (approved):

src/components/HdModal.vue
volcanioo: I understand what you meant thank you again man  https://github.com/homeday-de/homeday-blocks/pull/732/commits/dd0e5434436bc0ea4512a89baccfaf833bd3e061

ilyasmez: 🚀

PR #737, opened 2021-05-20
🐛  HD Modal Mobile Alignment Bugfix

volcanioo:
## WHAT
The bottom of the modal wasn't showing properly on mobile devices because of the browser's menu. We added a buffer to the padding of the modal to change the position a bit.

### DIFF
--- a/src/components/HdModal.vue
+++ b/src/components/HdModal.vue
@@ -167,6 +167,7 @@ $mobile-browser-footer-buffer: 75px;
   padding: #{$sp-l + $sp-s} $sp-m $sp-m;
   background-color: $white;
   border-radius: 4px;
+  overflow: hidden;
 }
 
 .hd-modal__header {
--- a/src/components/HdModal.vue
+++ b/src/components/HdModal.vue
@@ -131,6 +131,7 @@ $mobile-browser-footer-buffer: 75px;
 
 .hd-modal {
   position: fixed;
+  z-index: 100;
   display: flex;
   top: 0;
   left: 0;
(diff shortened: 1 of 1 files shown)


metal-gogo (approved):
Great catch!

ilyasmez:
Hey @volcanioo, this actually made me think of the case of a modal with a long content or the case of a screen with a short height (or in landscape orientation)... I think we should make the scrolling possible vertically. What do you think? Shall we include it in this PR or in a separate one?

volcanioo:
> Hey @volcanioo, this actually made me think of the case of a modal with a long content or the case of a screen with a short height (or in landscape orientation)... I think we should make the scrolling possible vertically. What do you think? Shall we include it in this PR or in a separate one?

I'm aware of this and planning to have another discussion with Magda about HD Modal's behavioral decisions, such as;

* Opening/Closing animations
* Mobile positioning as we experience calculation issues with browser's default behavior on Mobile
* Overlay and content sizing calculations
* I'll propose to use buttons in another way on mobile to reduce height. 

I want to do that in one time here, but I'm also trying to manage some FE tickets in my current sprint. :p I also don't want to make it one by one as these topics are essential for UX. What about tackling these issues in the next sprint with a TD ticket? In that way, I can find a time to talk to Magda to decide together on other issues too. What do you think @ilyasmez

metal-gogo:
Added a commit to include a `z-index` on the `hd-modal`: Without it on the customer app it looks like this:

Without `z-indez` | With `z-index`
--- | ---
Screen Shot 2021-05-20 at 23 19 25 | Screen Shot 2021-05-20 at 23 19 39

ilyasmez (approved):
Hey @volcanioo, as mentioned in our call this morning, we can fix the scrolling in a different PR.
🚀

aym3nb (approved):

PR #738, opened 2021-05-21
🐛  Scroll Bugfix of HD Modal

volcanioo:
## Bug

https://user-images.githubusercontent.com/14016028/119136228-b1957480-ba3f-11eb-8554-63c11b293dc3.MP4

## Fix
https://user-images.githubusercontent.com/14016028/119136494-020cd200-ba40-11eb-96e3-341f672f68c5.MP4

### DIFF
--- a/src/components/HdModal.vue
+++ b/src/components/HdModal.vue
@@ -168,7 +168,8 @@ $mobile-browser-footer-buffer: 75px;
   padding: #{$sp-l + $sp-s} $sp-m $sp-m;
   background-color: $white;
   border-radius: 4px;
-  overflow: hidden;
+  max-height: 90%;
+  overflow: scroll;
 }
 
 .hd-modal__header {


metal-gogo (approved):

src/components/HdModal.vue
```diff
@@ -168,7 +168,8 @@ $mobile-browser-footer-buffer: 75px;
   padding: #{$sp-l + $sp-s} $sp-m $sp-m;
   background-color: $white;
   border-radius: 4px;
-  overflow: hidden;
+  max-height: 90%;
```
metal-gogo, line 171: I think that 100% should work because the `hd-modal` class already takes into account the padding on the bottom for Android.

aym3nb (approved):

src/components/HdModal.vue
volcanioo: I will search about `env()` to fix this issue. Because 100% is not working properly on iOS Safari.

PR #744, opened 2021-06-18
👷 use slack-release-bot-action

viniciuskneves:
It uses the recently created GitHub Action to post releases on Slack: https://github.com/homeday-de/slack-release-bot-action

We have this `curl` in two places already (Sliders and Blocks) and it might be used by other projects so instead of having copy-paste everywhere we can have a single source of truth.

There is one difference: the action uses a `divider` between title and body.

### DIFF
--- a/.github/workflows/announce-release.yml
+++ b/.github/workflows/announce-release.yml
@@ -14,7 +14,8 @@ jobs:
         uses: LoveToKnow/slackify-markdown-action@v1.0.0
         with:
           text: ${{ github.event.release.body }}
-      - uses: homeday-de/slack-release-bot-action@0.0.1
+      - name: Announce on Slack 📢 
+        uses: homeday-de/slack-release-bot-action@0.0.1
         with:
           webhook_url: ${{ secrets.SLACK_RELEASE_BOT_WEBHOOK_URL }}
           title: ${{ github.event.release.name }}
--- a/.github/workflows/announce-release.yml
+++ b/.github/workflows/announce-release.yml
@@ -14,34 +14,9 @@ jobs:
         uses: LoveToKnow/slackify-markdown-action@v1.0.0
         with:
           text: ${{ github.event.release.body }}
-      - name: Announce on Slack 📢
-        run: |
-          curl \
-            -X POST \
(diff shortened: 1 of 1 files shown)


kuroski (approved):
This is soo cool, nice work on the GA bot

leandroinacio (approved):
🚀

ilyasmez (approved):
Much cleaner ✨ 


It will be nice if you could keep the step name, as it helps understanding the what that thing is about, and also looks good in the logs :D

I mean something like this:
```yml
 - name: Announce the release on the Frontend Slack channel 📢
   uses: homeday-de/slack-release-bot-action@0.0.1
   with:
     ....
```

.github/workflows/announce-release.yml
```diff
@@ -14,34 +14,9 @@ jobs:
...
-                  ]
-                }
-              ]
-            }' \
-            $SLACK_RELEASE_BOT_WEBHOOK_URL
-        env:
-          TITLE: ${{ github.event.release.name }}
-          NOTES: ${{ steps.release_body.outputs.text }}
-          SLACK_RELEASE_BOT_WEBHOOK_URL: ${{ secrets.SLACK_RELEASE_BOT_WEBHOOK_URL }}
+      - uses: homeday-de/slack-release-bot-action@0.0.1
```
viniciuskneves, line 17:
```suggestion
      - name: Announce on Slack 📢 
        uses: homeday-de/slack-release-bot-action@0.0.1
```

viniciuskneves:
> Much cleaner ✨
> 
> It will be nice if you could keep the step name, as it helps understanding the what that thing is about, and also looks good in the logs :D
> 
> I mean something like this:
> 
> ```yaml
>  - name: Announce the release on the Frontend Slack channel 📢
>    uses: homeday-de/slack-release-bot-action@0.0.1
>    with:
>      ....
> ```

@ilyasmez I just copied what we had before, let me know what you think: 75a72926391997c73c4f946e114f39ba59550b28

ilyasmez:
@viniciuskneves looks good! 👍  Thanks!

PR #748, opened 2021-06-22
✨ Modify loadScript function in order to take furtur group of attributes

alieslamifard:
Quite self-explanatory!
In case you need to set some more attributes on the script tag like below, just pass it as an object called props to the function.

How to use:
`loadScript({ url: '/example', props: { id: '1', 'data-whatever': '123' } })`

Output:
`<script src="/example" id="1" data-whatever="123" />`


leandroinacio (approved):
Congrats on your first pull request @alieslamifard! 🎉 
This looks great. 
One thing I would like to suggest is to update the documentation related to this function to add the new parameter there. But looks like we don't have any docs for it: https://github.com/homeday-de/homeday-blocks/tree/master/src/services#utils

It is not a blocker for your code commit, but feel free to create it :)

alieslamifard:
> Congrats on your first pull request @alieslamifard! 🎉
> This looks great.
> One thing I would like to suggest is to update the documentation related to this function to add the new parameter there. But looks like we don't have any docs for it: https://github.com/homeday-de/homeday-blocks/tree/master/src/services#utils
> 
> It is not a blocker for your code commit, but feel free to create it :)

Thanks @leandroinacio 💯 
I'll do it.

ilyasmez (commented):
Great! Thanks for taking care of that @alieslamifard

I have one remark: You're using `script[x] = y`, which sets the props of the element as you've mentioned, so in your example (`loadScript({ url: '/example', props: { id: '1', 'data-whatever': '123' } })`), only the `id` will show up as output, as it's a prop of `<script>` (the same for `src`, `type`...), but `data-whatever` won't, because it will be a hidden (unusable) prop. I think what you're aiming to achieve is to set attributes and not props, so I'd suggest to use `Element.setAttribute()` instead of assigning the props directly.
And also, it's always better to use an array function instead of a `for` loop, at least for readability reasons.
So the code snippet might look like this:
```js
Object.entries(attributes).forEach(([key, value]) => {
  script.setAttribute(key, value);
});
```

And we will have to rename the arguments `props` to `attributes`.

Let me know what you think 🙂

alieslamifard:
Thanks, @ilyasmez, LGTM.

viniciuskneves (commented):
Wanna cover it with tests? https://github.com/homeday-de/homeday-blocks/blob/master/tests/unit/services/utils.spec.js

I think JSDOM covers all the APIs used here.

ilyasmez (approved):
Looks good 👍 
Congrats on your first PR 🍫 

It will be nice if you could add test for this function as Vini mentioned, but this can be done on a different PR as well (just to not overwhelm you on your first PR 😄 ).

alieslamifard:
> Wanna cover it with tests? https://github.com/homeday-de/homeday-blocks/blob/master/tests/unit/services/utils.spec.js
> 
> I think JSDOM covers all the APIs used here.

Done!

package.json
```diff
@@ -75,7 +75,8 @@
     "semantic-release": "^17.4.4",
     "semantic-release-gitmoji": "^1.3.4",
     "vue-code-highlight": "^0.7.8",
-    "vue-template-compiler": "2.6.14"
+    "vue-template-compiler": "2.6.14",
+    "jsdom": "^16.6.0"
```
viniciuskneves, line 79: Jest uses JSDOM by default: https://jestjs.io/docs/26.x/configuration#testenvironment-string (unless we go Jest 27 which has new defaults but we are on 24 here).

It means that you don't need to install and import JSDOM by yourself. You can check this example from their docs: https://jestjs.io/docs/26.x/tutorial-jquery
It shows how it works with jQuery but it follows the same principles we need.

You can use `window` directly in the code as Jest will take care of setting it up for us (with JSDOM).

Let me know if you need help here!

alieslamifard: Done!

viniciuskneves (commented):
Hey @alieslamifard it looks good now but apparently something is wrong with `package-lock`. I think you had issues after installing/uninstalling jsdom, could you take a last look into it?

alieslamifard:
> Hey @alieslamifard it looks good now but apparently something is wrong with `package-lock`. I think you had issues after installing/uninstalling jsdom, could you take a last look into it?

I didn't uninstall it to avoid making problem with dependencies (since it has a peer dependency with jest). I just removed it's line in `package.json` and ran the `npm install` again.

viniciuskneves:
> > Hey @alieslamifard it looks good now but apparently something is wrong with `package-lock`. I think you had issues after installing/uninstalling jsdom, could you take a last look into it?
> 
> I didn't uninstall it to avoid making problem with dependencies (since it has a peer dependency with jest). I just removed it's line in `package.json` and ran the `npm install` again.
> 
> Update: I found the problem, that's because I ran the npm install on node `16.0.0`. I'll fix it.

To avoid such issue, I recommend: https://github.com/nvm-sh/nvm#deeper-shell-integration

viniciuskneves (approved):
My last comment about `package-lock` was the last one. Cool work here! Keep rocking 🚀

alieslamifard:
> > > Hey @alieslamifard it looks good now but apparently something is wrong with `package-lock`. I think you had issues after installing/uninstalling jsdom, could you take a last look into it?
> > 
> > 
> > I didn't uninstall it to avoid making problem with dependencies (since it has a peer dependency with jest). I just removed it's line in `package.json` and ran the `npm install` again.
> > Update: I found the problem, that's because I ran the npm install on node `16.0.0`. I'll fix it.
> 
> To avoid such issue, I recommend: https://github.com/nvm-sh/nvm#deeper-shell-integration

Nice, Worked like a charm!

PR #753, opened 2021-06-23
👷 missing 'v' in version

viniciuskneves:
You can kill me now or kill me later, doesn't matter 🙈

### DIFF
--- a/.github/workflows/announce-release.yml
+++ b/.github/workflows/announce-release.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           text: ${{ github.event.release.body }}
       - name: Announce on Slack 📢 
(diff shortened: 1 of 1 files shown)


leandroinacio (approved):

ilyasmez (commented):
🤔 ... maybe later, we still need you.

ilyasmez (approved):

ilyasmez:
One little info, this is not a bug in the library (the end user/developer is not affected), but in the CI.... so the gitmoji 🐛  is not so accurate here as it will have a semantic versioning effect... so you can just use 👷 or similar gitmojis (you can see the `semver` of every gitmoji [here](https://github.com/carloscuesta/gitmoji/blob/master/src/data/gitmojis.json)).

PR #782, opened 2021-07-14
✨[DS-166]  - Radio and Checkbox card components

kuroski:
https://homeday.atlassian.net/browse/DS-166

In this PR:
- Change our jest config file to actually use `vue-cli` preset
- Improve a few files for autocompletion
  - jsconfig module resolution
  - vue shims
  - jsdocs in language service
  - lodash definitely typed definitions dependency
- Add Radio and Checkbox indicator components
- Add Radio and Checkbox 
  - card components
  - group components

❗ 
> If possible, check the PR through the commit history 😉 
> If the component selection is not working, try refresh the storybook, I faced a few inconsistencies with `v-model` when changing from "Canvas" to "Docs" tab

The main goal of this PR is to introduce the new components
- HdRadioCard
- HdRadioCardGroup
- HdCheckbox
- HdCheckboxGroup

Those components will be initially used in our new myHomeday screen:
image

They have a very similar API, you can use them in the following way:

``` HTML
<!-- HdRadioCard -->
<HdForm @submit="onSubmit">
  <HdRadioCard name="property" native-value="apartment" v-model="propertyType">
    <template #icon>
      <HdIcon :src="apartmentCommercialIcon" />
    </template>

    Apartment
  </HdRadioCard>

  <HdRadioCard name="property" native-value="castle" v-model="propertyType">
    <template #icon>
      <HdIcon :src="houseCastleIcon" />
    </template>

    Castle
  </HdRadioCard>
</HdForm>

<!-- With HdRadioCardGroup -->
<HdForm @submit="onSubmit">
  <HdRadioCardGroup class="radio-group" name="property-group" v-model="propertyType">
    <HdRadioCard native-value="apartment">
      <template #icon>
        <HdIcon :src="apartmentCommercialIcon" />
      </template>

      Apartment
    </HdRadioCard>

    <HdRadioCard native-value="castle">
      <template #icon>
        <HdIcon :src="houseCastleIcon" />
      </template>

      Castle
    </HdRadioCard>
  </HdRadioCardGroup>
</HdForm>
```

### They will replace the card components we have in sliders?
img

Well, for now, no... those components were created initially to supply a new feature in the exposé preparation module.

### Why `HdRadioIncidator` and `HdCheckboxIndicator`?

I've made a planning session before starting this ticket, and after talking with @ilyasmez we ended up deciding on actually creating those components. 

We followed this approach to avoid having breaking changes in the code since `HdCheckbox` works with single elements and `HdRadio` works with grouped elements.

This is a bit tricky since those new components are supposed to work alone and with groups.

Extracting those indicators helped to reduce a bit of code on `HdRadio` and `HdCheckbox` components, we were able to introduce NO breaking changes, and they would fit in the scope of the new components.

### Why I didn't extract common logic from the new components?

- They are pretty similar, but not the same, they have small changes between them, and I was a bit afraid of doing some wrong abstraction here
- I probably could wrap up some mixings and share a few things, but maybe it worth planning a bit better
- I ended up working on this task for a while now, so I want to actually have this first version on, and then work in other reactors (including `HdRadio` and `HdCheckbox` in this scope)


ilyasmez (approved):
Looooooooove it!! 😃 
Fantastic work! 🙇 🎩 👏 

And for the rest of the team, just to avoid confusion, if you have more than one RadioCard or CheckboxCard, then you have to use the Group component. One of the reasons is that you'll have this:
✅
Screen Shot 2021-07-14 at 10 39 48 AM
instead of this:
❌
Screen Shot 2021-07-14 at 10 39 31 AM

____

Just one tiny little minor note... you're using `outline` to show the focus indicator... it makes complete sense, it only doesn't stick to the shape of the element itself, it covers just the bounding box, so the effect looks like this:
Screen Shot 2021-07-14 at 9 22 31 AM
You can double check it with the design team, and we can think of using box-shadow or an ::after/::before pseudo element. But of course, this PR is more than ready to go live... such a minor adjustment can be made in a new PR 🙂

tests/unit/components/buttons/__snapshots__/HdButton.spec.js.snap
```diff
@@ -51,7 +51,7 @@ exports[`HdButton should render component with btn--tertiary class 1`] = `
 
 exports[`HdButton should render the icon passed as prop 1`] = `
 <button class="btn btn--primary"><span class="btn__icon"><svg height="100%" width="100%" viewBox="222 126 53 53" version="1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns="http://www.w3.org/2000/svg">
-  <path class="color1" d="M272.48 137.62L258.61 150.5L244.75 141.58L254.65 126.73L272.48 137.62Z" fill="#1895FF"></path>
-  <path class="color2" d="M242.22 176.24L222.97 160.68L230.47 145.1L243.56 155.68L242.22 176.24Z" fill="#1C3553"></path>
+  <path xmlns="http://www.w3.org/2000/svg" class="color1" d="M272.48 137.62L258.61 150.5L244.75 141.58L254.65 126.73L272.48 137.62Z" fill="#1895FF"></path>
```
volcanioo, line 54: Hey man, just out of curiosity, is there any specific reason to add `xmlns="http://www.w3.org/2000/svg"` to the path?

src/components/form/HdRadioIndicator.vue
```diff
@@ -0,0 +1,118 @@
...
+      type: Boolean,
+      default: false,
+    },
+    invalid: {
+      type: Boolean,
+      default: false,
+    },
+  },
+};
+</script>
```
volcanioo, line 19: Is there any specific reason to start with `script` instead of `template`? I like it btw =D

tests/unit/components/__snapshots__/HdIcon.spec.js.snap
```diff
@@ -2,16 +2,16 @@
 
 exports[`HdIcon adds title and desc tag to the svg 1`] = `
 <svg aria-describedby="icon-id-desc" height="50" width="50" viewBox="222 126 53 53" version="1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns="http://www.w3.org/2000/svg" id="icon-id">
-  <path class="color1" d="M272.48 137.62L258.61 150.5L244.75 141.58L254.65 126.73L272.48 137.62Z" fill="#1895FF"></path>
-  <path class="color2" d="M242.22 176.24L222.97 160.68L230.47 145.1L243.56 155.68L242.22 176.24Z" fill="#1C3553"></path>
-  <desc id="icon-id-desc">Some description.</desc>
-  <title>This make me accessible.</title>
+  <path xmlns="http://www.w3.org/2000/svg" class="color1" d="M272.48 137.62L258.61 150.5L244.75 141.58L254.65 126.73L272.48 137.62Z" fill="#1895FF"></path>
```
volcanioo, line 5: Is there any spesific reason to add `xmlns="http://www.w3.org/2000/svg"` to the `path` element?

volcanioo (approved):
Great commit history and clear work. 🚀  Thanks for that.

kuroski:
Thanks a lot for your comment @ilyasmez 
Alongside with Magda and Anna's comments, I've made a few fixes in the styling:

- 4b48267 

I just didn't enjoyed much the svg fill option for disable state.

ilyasmez:
> Alongside with Magda and Anna's comments, I've made a few fixes in the styling:
> * [4b48267](https://github.com/homeday-de/homeday-blocks/commit/4b48267c54a5b18b1b5c17b25098e062c5e3beeb)
> I just didn't enjoyed much the svg fill option for disable state.

Hey @kuroski, it's hacky indeed, but no worries, this is something we are planning to work on, HdIcon already offers a `fillToClass` property which can be used here (so we can have better CSS selectors), and we can also have a high level API, like add a `disabled` property to HdIcon.
But for now we can go with the hack and revisit this when we agree on a solution.

tests/unit/components/buttons/__snapshots__/HdButton.spec.js.snap
kuroski: Soo, this was just changed for the testing scenario, this is something related to this commit:

099b7f3 

Now we are using the preset coming from `vue-cli` for our tests, and behind the scenes, they are adding that to SVG parsing, but in the actual implementation, things are normal.

src/components/form/HdCheckboxCard.vue
```diff
@@ -0,0 +1,315 @@
+<template>
+  <label
+    ref="label"
+    :for="inputName"
+    @click="$refs.checkbox.focus()"
+    @keydown.prevent.enter="$refs.label.click()"
```
metal-gogo, line 6: I don't think that we should be listening to a `enter` key. On a [regular checkbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox) the enter key does not work, only the `space` key... Usually we should use the checkboxes inside a form, and, in general, the enter should trigger the form submit. What do you think?

src/components/form/HdRadioIndicator.vue
kuroski: Oh my, thanks for spotting that, I didn't notice that, I have a boilerplate in my code editor, but it was created with `scripts` in the first place, I already changed that, since we are using `templates` first 

e69e869

src/components/form/HdCheckboxCard.vue
kuroski: Thanks a lot, man, you are totally right =D
4fdfdd6

kuroski:
Hey @ilyasmez thanks a lot for the suggestion, the filter option is waaaay better compared to what I did.
I already had a sync with the designers and everything seems to be fine to use, I combined the filter with opacity and we achieved a mid term behaviour:

439d7ba

ilyasmez:
> Hey @ilyasmez thanks a lot for the suggestion, the filter option is waaaay better compared to what I did.
> I already had a sync with the designers and everything seems to be fine to use, I combined the filter with opacity and we achieved a mid term behaviour:
> 
> [439d7ba](https://github.com/homeday-de/homeday-blocks/commit/439d7ba21674590c60b9a0b70be44dc36b23bd81)

Cool! Thank you for the effort! Both solutions are temporary, so it's nice that we went for the simplest :)

And btw (super minor), I think you can just apply the filter to the icon element without targeting, like so:
```scss
.card--disabled > .card__icon {
  filter: grayscale(0.95);
  opacity: 0.4;
}
```
(didn't test so take it with some grain of salt 😄)

src/components/form/HdCheckboxIndicator.vue
```diff
@@ -0,0 +1,156 @@
...
+      return null;
+    },
+  },
+};
+</script>
+
+<style lang="scss" scoped>
+@import "homeday-blocks/src/styles/mixins.scss";
+
+.indicator {
```
metal-gogo, line 58: I propose that we remove the `scoped` keyword. This could create an issue because `indicator` is kind of generic, but maybe we should make the class more specific? Perhaps `checkbox-indicator`?

src/components/form/HdRadioCard.vue
```diff
@@ -0,0 +1,280 @@
...
+
+<style lang="scss" scoped>
+@import "homeday-blocks/src/styles/mixins.scss";
+
+::v-deep path {
+  transition: fill ($time-s * 2);
+}
+
+input[type="radio"] {
+  display: none;
```
metal-gogo, line 172: This would remove the input from the tab order. Is that what we want? Maybe we want to keep the tab order in place and just make it visually hidden?
We could even create a mixin for this:

```scss
/// Module export mixin
/// This mixin helps creating a practically hidden element. It's important to note that this
/// mixin will hide the element for the user, but will be kept available to screen readers
/// and other assistive technology users. They are hidden with CSS, but they remain on the DOM.
@mixin visually-hidden() {
  position: absolute;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  clip: rect(0 0 0 0);
  border: 0;
}
```

src/components/form/HdRadioCard.vue
```diff
@@ -0,0 +1,280 @@
...
+
+.card--tile > .card__control-label {
+  flex: 1;
+  margin-top: 0;
+  margin-left: $sp-xs;
+  margin-right: $sp-xs;
+}
+
+.card--tile ::v-deep svg {
+  max-width: 44px;
```
metal-gogo, line 270: Should this be relative to the font size? We have a `rem-calc` function if it is needed

src/components/form/HdRadioIndicator.vue
```diff
@@ -0,0 +1,118 @@
...
+
+<style lang="scss" scoped>
+@import "homeday-blocks/src/styles/mixins.scss";
+
+.indicator {
```
metal-gogo, line 35: Same here, I propose that we remove the `scoped` keyword. This could create an issue because `indicator` is kind of generic, but maybe we should make the class more specific? Perhaps `radio-indicator`?

src/components/form/HdRadioCard.vue
kuroski: Well, this is not what we want, we already have the checkbox/radio surrounded by the label, if we use this snippet in this case, when the user is navigating through the form, they will have to hit the tab twice to select the next component.

Check the example below, I'm only navigating through the keyboard, and every tab I hit, it selects first the checkbox (that doesn't display any visual feedback), and when I hit again it selects the label.


https://user-images.githubusercontent.com/1855125/125812357-e29dbd1a-c5d0-4afc-ad57-67c5e97346ff.mov

What do you think about this?

kuroski:
Thanks a lot @ilyasmez, it worked 😄 

83a5dd5

src/components/form/HdRadioIndicator.vue
kuroski: Thanks for your suggestion man =D here are the changes

9bb0eee

src/components/form/HdCheckboxIndicator.vue
kuroski: Thanks for your suggestion man =D here are the changes

9bb0eee

src/components/form/HdRadioCard.vue
kuroski: That's a good catch!

5d1a72c

metal-gogo (approved):
Great work man!!

PR #783, opened 2021-07-20
✨Modify HdModal to take a root element in order to have a fully custom modal

alieslamifard:
Basically, HdModal has header, body, footer, and action buttons with a specific style, but there might be some cases that we need to have a fully customizable modal (e.g. sliders modal).

Here is how it works:
```vue
<button @click="showModal">show modal</button>
<hd-modal
  v-if="isOpen"
  is-clean-sheet="true"
>
  <template slot="root">
    <div class="custom-root-element">
      <p>This is a modal with <b> custom root element</b>.</p>
      <button class="button" @click="hideModal">Custom Hide Button</button>
    </div>
  </template>
</hd-modal>
```


drapisarda:
what about having `isClearSheet` as a computed with
```js
isClearSheet() {
   return !!this.$slots.root;
}
```

?

src/components/HdModal.vue
```diff
@@ -6,7 +6,7 @@
       :class="{ 'hd-modal--with-icon': withIcon }"
     >
       <div class="hd-modal__overlay" />
-      <div class="hd-modal__container">
+      <div class="hd-modal__container" v-if="!isCleanSheet">
```
volcanioo, line 9: Shouldn't we put `v-if` right before the `class` attribute?

volcanioo (approved):
Hey Ali, great work. I just posted one small optional detail. 🚀 🚀

viniciuskneves:
> what about having `isClearSheet` as a computed with
> 
> ```js
> isClearSheet() {
>    return !!this.$slots.root;
> }
> ```
> 
> ?

I would agree with something in that direction as well. I don't get why we need two properties to make it work: slot called `root` + prop.

Is it really necessary to create a new slot? Is `root` a standard naming for this case too?

alieslamifard:
> > what about having `isClearSheet` as a computed with
> > ```js
> > isClearSheet() {
> >    return !!this.$slots.root;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> I would agree with something in that direction as well. I don't get why we need two properties to make it work: slot called `root` + prop.
> 
> Is it really necessary to create a new slot? Is `root` a standard naming for this case too?

I removed the prop.
For naming part, we already have body, header, footer inside some elements, so I decided to have a quite straightforward name for this slot. Although we can have something like content either. 
Let me know what do you think Vini.

src/components/HdModal.vue
alieslamifard: Sure. Done.

leandroinacio (approved):
Looks good to me 🚀

alieslamifard:
Also renamed `isCleanSheet` to `hasCustomContent` for the sake of simplicity.

viniciuskneves:
I will sleep on it hahaha, if it gets approved beforehand I drop some comments for improvements later.

My main concern is: how do I know that defining `root` slot means that all other slots are ignored and the "default styles" won't be applied.

alieslamifard:
> I will sleep on it hahaha, if it gets approved beforehand I drop some comments for improvements later.
> 
> My main concern is: how do I know that defining `root` slot means that all other slots are ignored and the "default styles" won't be applied.

Good catch! We can go for something like: `replace-everything` or `content-less` or `cleanSheet`. Funny names BTW😆.
Nevertheless, I tried to make it understandable in its storybook.

kuroski:
Well, that was a fast PR 😆  
But actually, we could have just worked with default slots overrides:

https://codesandbox.io/embed/vue-nested-slots-30hgg?autoresize=1&fontsize=14&hidenavigation=1&theme=dark

You don't need to provide a prop, just use the defaults.

But that's not a problem.

alieslamifard:
> Well, that was a fast PR 😆
> But actually, we could have just worked with default slots overrides:
> 
> https://codesandbox.io/embed/vue-nested-slots-30hgg?autoresize=1&fontsize=14&hidenavigation=1&theme=dark
> 
> You don't need to provide a prop, just use the defaults.
> 
> But that's not a problem.

Nice, Thanks for the examples, but we have that **_overlay div_** and what to keep it.

PR #785, opened 2021-07-22
✨ Add closing event on HdModal Overlay

alieslamifard:
### Modal should be closed when a user clicks on the overlay.
----

This is the outcome of the last [Design system sync](https://homeday.atlassian.net/wiki/spaces/PE/pages/2209218870/2021-07-21+Design+System+Inactive+icons+focus+toast). so clicking on overlay should close the HdModal.

> Our modal definitely needs to be updated so that it automatically closes when the user clicks anywhere outside of it.

### DIFF
--- a/src/components/HdModal.vue
+++ b/src/components/HdModal.vue
@@ -5,7 +5,7 @@
       class="hd-modal"
       :class="{ 'hd-modal--with-icon': withIcon }"
     >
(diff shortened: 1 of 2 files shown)


volcanioo (approved):

leandroinacio (approved):
Thanks for taking care of that 👍

PR #793, opened 2021-07-28
✨ Add keysToCamelCase & keysToSnakeCase helper function into UtilsService

peterchencc:
https://homeday.atlassian.net/browse/VIEW-291

We are using `keysToCamelCase` & `keysToSnakeCase` helper functions in customer-app and Sliders. 
As @viniciuskneves 's [suggestion](https://github.com/homeday-de/sliders/pull/1700#pullrequestreview-714678698)
It makes sense to move them to Blocks and have a single source of truth.

Reference from customer-app: 
- [src/services/utils.js#L105-L141](https://github.com/homeday-de/customer-app/blob/develop/src/services/utils.js#L105-L141)
- [tests/unit/services/utils.spec.js#L323-L333](https://github.com/homeday-de/customer-app/blob/develop/tests/unit/services/utils.spec.js#L323-L333)

Will create a tech debt ticket to migrate those current functions to use the helper functions from Blocks in customer-app. 
Of course, after we upgrade homeday-blocks to the newest version in customer-app :)

### DIFF
--- a/src/services/utils.js
+++ b/src/services/utils.js
@@ -171,6 +171,7 @@ export function cloneVNodeElement (vnode, { props, attrs, children, ...rest }, h
   }, cloned.componentOptions.children || children)
 }
 
+// Deeply converts the keys of a given object to camelCase
 export function keysToCamelCase(collection) {
   if (_isPlainObject(collection)) {
     const newCollection = {};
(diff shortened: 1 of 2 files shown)


viniciuskneves (approved):
Really cool! Loved the tests as well 😄

leandroinacio (approved):
🚀

PR #796, opened 2021-08-03
🐛 Add main.scss file to package.json sideEffects prop

kuroski:
After the [previous release](https://github.com/homeday-de/homeday-blocks/pull/794) we had a problem with global styling inside Storybooks:

image

To fix that, we added this property to include `main.scss` file into blocks context.

## Notes

- This is not actually a good thing since every component should contain its own styles and not depend on global/external styles, and we should at some point in time refactor that
- For now, we added this "quick fix" to always include the global styling in blocks, but the tree shaking will continue to work in other applications
- We also added one extra instruction step inside our "Welcome" story to warn the user that they must include the global styling

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   },
   "main": "main.js",
   "module": "main.js",
-  "sideEffects": ["*.vue"],
+  "sideEffects": ["*.vue", "src/styles/main.scss"],
   "files": [
     "src/*",
     "*.json",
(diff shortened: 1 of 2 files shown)


alieslamifard (approved):
Thanks for taking care of this @kuroski 💯

aym3nb (approved):
Good catch 👀 👍

drapisarda (approved):

PR #806, opened 2021-08-24
:lipstick: remove fixed width of HdGalleryCarousel image

REX-Alexander:
[Ticket: REAL-4009](https://homeday.atlassian.net/secure/RapidBoard.jspa?rapidView=7&modal=detail&selectedIssue=REAL-4009)

**Before:** 

before

**After removing `width:100%` from `HdGalleryCarousel image`:** 

after

### DIFF
--- a/src/components/gallery/HdGalleryCarousel.vue
+++ b/src/components/gallery/HdGalleryCarousel.vue
@@ -294,7 +294,6 @@ export default {
       z-index: 1;
 
     img {
-      width: 100%;
       height: 100%;
       object-fit: cover;
     }


src/components/gallery/HdGalleryCarousel.vue
```diff
@@ -294,7 +294,6 @@ export default {
       z-index: 1;
 
     img {
-      width: 100%;
       height: 100%;
       object-fit: cover;
     }
```
volcanioo, line 299: Hey mate! Nice approach! Removing the `width` property works, but what do you think using `object-fit: contain` instead of `object-fit: cover` to fix the issue? https://www.loom.com/i/575ff9863d0546e8ac08518997619f59 - Because `object-fit: cover` is the main reason the image looks covered within the element.

Nevertheless, I'm okay with your approach as well if removing the `width` makes more sense to you! But, shouldn't we remove the `object-fit` property completely as it becomes unnecessary once we remove the `width` property? Here is the same result without having `width` & `object-fit`. https://www.loom.com/i/7f2d23416c9d462db6a933247d42ae8b

ilyasmez (commented):
Hey @REX-Alexander! Congrats on your first PR to homeday-blocks!! 🎉  🥳 

I +1 @volcanioo's comment here. I think that many would expect the carousel (which is mainly used for the gallery's thumbnails) to have a "cover" style, and changing that might give them undesired UI.

I would suggest to allow passing the value of `object-fit` as a prop (it can default to `cover`).
Let me know what you think, and if you'd like to pair on this one 🙂

PR #815, opened 2021-09-06
✨ [VIEW-209]: Maxlength support for hdtextarea

volcanioo:
Ticket: https://homeday.atlassian.net/browse/VIEW-209

### DIFF
--- a/src/components/form/HdTextarea.vue
+++ b/src/components/form/HdTextarea.vue
@@ -92,7 +92,7 @@ export default {
     },
     maxlength: {
       type: Number,
-      default: () => Number.POSITIVE_INFINITY,
+      default: Number.POSITIVE_INFINITY,
     },
     lang: {
       type: String,
--- a/src/components/form/HdTextarea.vue
+++ b/src/components/form/HdTextarea.vue
@@ -12,7 +12,7 @@
     @clear-click="clearTextarea"
     @status-click="focusTextarea"
     :class="{
-      'field--maxlength-indicator-visible': Boolean(maxlength)
+      'field--maxlength-indicator-visible': isMaxlengthIndicatorVisible
     }"
   >
     <textarea
@@ -34,10 +34,10 @@
       @input="handleInput"
     />
     <span
-      v-if="maxlength"
+      v-if="isMaxlengthIndicatorVisible"
       class="maxlength-indicator"
     >
-      {{ value.length }}/{{maxlength}}
+      {{ maxlengthIndicator }}
     </span>
     <!-- `data-gramm_editor` attribute is used to control Grammarly (chrome extension) -->
   </TextFieldBase>
@@ -91,8 +91,8 @@ export default {
       default: '100px',
     },
     maxlength: {
-      type: [String, Number],
-      default: '',
+      type: Number,
+      default: () => Number.POSITIVE_INFINITY,
     },
     lang: {
       type: String,
@@ -125,6 +125,12 @@ export default {
     },
     isFilled() {
       return this.value !== null && this.value !== undefined && this.value !== '';
+    },
+    maxlengthIndicator() {
+      return `${this.value.length}/${this.maxlength}`;
+    },
+    isMaxlengthIndicatorVisible() {
+      return this.maxlength < Number.POSITIVE_INFINITY;
     },
     placeholderAttr() {
       if (!this.placeholder) {
@@ -197,7 +203,7 @@ export default {
   bottom: -#{($sp-s * 2) + ($sp-xs / 2)};
   @include font("DS-60");
   padding: 0px $sp-s;
-  color: getShade($neutral-gray, 70);
+  color: getShade($quaternary-color, 70);
 }
 .field--maxlength-indicator-visible {
   ::v-deep {
--- a/src/components/form/HdTextarea.vue
+++ b/src/components/form/HdTextarea.vue
@@ -192,6 +192,7 @@ export default {
 
 <style scoped lang="scss">
 @import 'homeday-blocks/src/styles/mixins.scss';
+$distanceFromTextareaToIndicator: -18px;
 
 .textarea {
   display: block;
@@ -200,7 +201,7 @@ export default {
 .maxlength-indicator {
   position: absolute;
   right: 0;
-  bottom: -#{($sp-s * 2) + ($sp-xs / 2)};
+  bottom: $distanceFromTextareaToIndicator;
   @include font("DS-60");
   padding: 0px $sp-s;
   color: getShade($quaternary-color, 70);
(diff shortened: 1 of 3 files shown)


src/components/form/HdTextarea.vue
```diff
@@ -11,6 +11,9 @@
     :disabled="disabled"
     @clear-click="clearTextarea"
     @status-click="focusTextarea"
+    :class="{
+      'field--maxlength-indicator-visible': Boolean(maxlength)
```
ilyasmez, line 15: Here you're using `Boolean(maxlength)` and later you use `v-if="maxlength"`, they are both the same, you can just stick to one to be consistent.

src/components/form/HdTextarea.vue
```diff
@@ -80,6 +90,10 @@ export default {
       type: String,
       default: '100px',
     },
+    maxlength: {
+      type: [String, Number],
+      default: '',
```
ilyasmez, line 95: Why do you allow `String`? To be able to hide the maxlength indicator? You're using `Boolean(maxlength)` which will resolve to `false` even if the maxlength is `0`. So no need to allow `String` as a type.

But I would suggest to use `Number.POSITIVE_INFINITY` as a default value, and have a computed property `isMaxlengthIndicatorVisible` that looks something like this:
```js
isMaxlengthIndicatorVisible() {
 return this.maxlength < Number.POSITIVE_INFINITY;
}
```

src/components/form/HdTextarea.vue
```diff
@@ -171,8 +185,25 @@ export default {
...
   display: block;
   resize: vertical;
 }
+.maxlength-indicator {
+  position: absolute;
+  right: 0;
+  bottom: -#{($sp-s * 2) + ($sp-xs / 2)};
+  @include font("DS-60");
+  padding: 0px $sp-s;
+  color: getShade($neutral-gray, 70);
```
ilyasmez, line 200:
```suggestion
  color: getShade($quaternary-color, 70);
```

src/stories/HdTextarea.stories.js
```diff
@@ -13,6 +13,7 @@ storiesOf('Components/Form/HdTextarea', module)
       <HdTextarea
         v-model="value"
         :required="true"
+        :maxlength="500"
```
ilyasmez, line 16:
```suggestion
```

It might seem like a default behavior, it's better if we keep it in its own story.

ilyasmez (commented):
Hey @volcanioo! The feature looks good!
Please check my comments and let me know what you think.

We should also update the design team with this change so they can document it on Zeroheight as well.

src/components/form/HdTextarea.vue
volcanioo: oh, i forgot about that! :d thanks!

src/components/form/HdTextarea.vue
volcanioo: Got your point; thanks for the idea; here are the changes: https://github.com/homeday-de/homeday-blocks/pull/815/commits/504d02ceac39fc5ee2fdceb9ef2a76471a98a41c

> Why do you allow String?
I was allowing String to avoid console errors but can't see the errors now. There might be a case where we pass a number as a string value from a JSON object through the component directly. The reason was to avoid the console errors in these specific cases, but I can't produce them atm.

src/components/form/HdTextarea.vue
volcanioo: nice catch, i started to use `isMaxlengthIndicatorVisible` variable instead

src/components/form/HdTextarea.vue
volcanioo: https://github.com/homeday-de/homeday-blocks/pull/815/commits/504d02ceac39fc5ee2fdceb9ef2a76471a98a41c

src/stories/HdTextarea.stories.js
volcanioo: oh right, i already created another case for that. you can see the changes here: https://github.com/homeday-de/homeday-blocks/pull/815/commits/504d02ceac39fc5ee2fdceb9ef2a76471a98a41c

src/components/form/HdTextarea.vue
```diff
@@ -80,6 +90,10 @@ export default {
       type: String,
       default: '100px',
     },
+    maxlength: {
+      type: Number,
+      default: () => Number.POSITIVE_INFINITY,
```
ilyasmez, line 95: Couldn't you just use the value directly instead of a factory function?

```suggestion
      default: Number.POSITIVE_INFINITY,
```

volcanioo: https://github.com/homeday-de/homeday-blocks/pull/815/commits/709415abb180a5a8b4e86abcf78dd008768fc645

src/components/form/HdTextarea.vue
```diff
@@ -171,8 +191,25 @@ export default {
...
+@import 'homeday-blocks/src/styles/mixins.scss';
+
 .textarea {
   display: block;
   resize: vertical;
 }
+.maxlength-indicator {
+  position: absolute;
+  right: 0;
+  bottom: -#{($sp-s * 2) + ($sp-xs / 2)};
```
kuroski, line 203: I understand that we have to add a specific amount of spacing from the counter to the text area.

But why do we have those operations?

I'm asking that, because we should only use what we have mapped in the DS without the need to do things like `sp-s + sp-xs`... for example, doing that sum will result in `12px` which is not present in the DS.

If you have to do that to conform to the design spec, we should then talk with the design team to either adapt the spec to what we have in the DS or add a new spacing to it.

ilyasmez: +1
You don't always have to force the usage of `$sp` values. Unless it's really needed (usually when using absolute/fixed positioning, where you want to add the padding of the container + the margin that you want for your element), but that should stay just for edge cases.
This one looks very complicated and hard to understand too, I would go for creating a variable with a semantic name and give it a hardcoded value for now (e.g. `$distanceFromAtoB: 99px;`), and we can discuss this use case in the DS sync if needed. You can also see if you can combine `bottom` and `margin-bottom` to achieve a more reasonable implementation.

aym3nb: I also suggest writing the operations without wrapping them in SASS `#{..}` because if I remember correctly it doesn't work in IE11 😅

aym3nb (approved):
👏 🚀

src/components/form/HdTextarea.vue
volcanioo: ℹ️  I have a problem on localhost, will push the changes in a couple of hours.

ilyasmez (approved):
Looks great! Thank you 🌋

drapisarda (approved):

aym3nb (approved):

PR #816, opened 2021-09-07
💡 Suggestion about the carousel component

leandroinacio:
Hey @homeday-de/frontend-engineering, I don't have any intention to merge this (yet) but to raise a question and see everyone's thoughts.
About the HdGalleryCarousel component.

**Current state**
It's a hidden component (we don't have stories for it - so no documentation) that we have in homeday-blocks and allows us to create a carousel for images. Like the gif below.
Components _ Gallery _ HdGalleryCarousel - Default ⋅ Storybook

Right now it is being used on the HdGallery, also located in homeday-blocks. So we need to make sure this continue working as it is right now.

**Suggestion**
Why don't we make it more reusable? On the website migration we have the need for a component like this:
image

One way to achieve it is to have a slot where we currently have the .gallery-carousel__item, rename the component to HdCarousel and then on the new HdGalleryCarousel we can just fill in the slot with the .gallery-carousel__item (as it was before). So the current component would continue working as it is right now and open the possibility to add something else in the carousel besides just images.

The code below contemplates just part of the suggested changes, but it's enough to get a preview:
Components _ Gallery _ HdGalleryCarousel - Slot ⋅ Storybook

This would enable the articles component on our website coming from this component, and we would still have the old component working.

What do you think? Let me know.

### DIFF
--- a/src/components/gallery/HdGalleryCarousel.vue
+++ b/src/components/gallery/HdGalleryCarousel.vue
@@ -14,26 +14,28 @@
         ref="flickity"
         :options="flickityOptions"
       >
-        <div
-          class="gallery-carousel__item"
-          v-for="(item, i) in items"
-          :key="i"
-          :class="{
-            'is-active': shouldShowActiveState(i),
-          } "
-        >
-          <div :style="sizerStyles" />
+        <slot>
+          <div
+            class="gallery-carousel__item"
+            v-for="(item, i) in items"
+            :key="i"
+            :class="{
+              'is-active': shouldShowActiveState(i),
+            } "
+          >
+            <div :style="sizerStyles" />
 
-          <!-- the item.thumbnail field is used as default value for the item image -->
-          <!-- IE11 uses this value only because do not support the picture element -->
-          <picture class="gallery-carousel__picture">
-            <source v-for="(source, media) in item.thumbnailPictureSources"
-                    :key="media"
-                    :media="`(${media})`" :srcset="source"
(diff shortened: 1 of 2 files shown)


ilyasmez:
Hey @leandroinacio! Thank you for opening this discussion.

First, I do agree that `HdGalleryCarousel` should have its own stories.
Regarding reusing it for other purposes, I'm not sure about that. From the examples you mentioned, I think it would make much more sense to have a generic `HdCarousel` which takes care of the carousel effect and having a pager. We can then think of reusing this new component (`HdCarousel`) inside `HdGalleryCarousel`, it will be an `HdCarousel` + all the gallery specific features, such as zoom on click and so on.

drapisarda:
> Hey @leandroinacio! Thank you for opening this discussion.
> 
> First, I do agree that `HdGalleryCarousel` should have its own stories.
> Regarding reusing it for other purposes, I'm not sure about that. From the examples you mentioned, I think it would make much more sense to have a generic `HdCarousel` which takes care of the carousel effect and having a pager. We can then think of reusing this new component (`HdCarousel`) inside `HdGalleryCarousel`, it will be an `HdCarousel` + all the gallery specific features, such as zoom on click and so on.

So basically what you did but in a separate component and therefore with not default content for the slot. I do agree.

leandroinacio:
Created this ticket, so I'm closing the draft: https://homeday.atlassian.net/browse/FET-247

PR #830, opened 2021-10-05
✨ HdButton doesn't inherit attributes

viniciuskneves:
It disables attribute inheritance from HdButton.
In the website project we've an use case with a default where the HdButton is inheriting the attributes. Is it a big deal? No.

I would leave it up to you to check if you are using attribute inheritance somewhere in your projects. I will wait for an approval for Customer App and from Realtor App at least.

---

_package-lock.json and package.json changes were the outcome of `npm i`_

### DIFF
--- a/tests/unit/components/buttons/HdButton.spec.js
+++ b/tests/unit/components/buttons/HdButton.spec.js
@@ -101,4 +101,14 @@ describe('HdButton', () => {
 
     expect(click).toHaveBeenCalled();
   });
+
+  it('does not inherit attributes', async () => {
+    const wrapper = wrapperBuilder({
+      props: {
+        myFakeProp: 'HUEBR',
+      },
+    });
+
+    expect(wrapper.attributes('myfakeprop')).toBeUndefined();
+  });
 });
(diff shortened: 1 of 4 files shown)


tests/unit/components/buttons/HdButton.spec.js
```diff
@@ -101,4 +101,14 @@ describe('HdButton', () => {
 
     expect(click).toHaveBeenCalled();
   });
+
+  it('does not inherit attributes', async () => {
+    const wrapper = wrapperBuilder({
+      props: {
+        myFakeProp: 'HUEBR',
```
leandroinacio, line 108: 😂

ilyasmez (approved):
Great! This should be the case for all the "transparent components"

leandroinacio (approved):
This kind of thing will also happen on HdLink.

tests/unit/components/buttons/HdButton.spec.js
viniciuskneves: image

volcanioo: dude hahahahhaah thats great :D

volcanioo (approved):

PR #856, opened 2021-10-29
🐛 Make the gallery components styles scoped

leandroinacio:
@viniciuskneves' eagle eyes just saw that we had some CSS leaking into our Carousel on the website project.
Since all our components are using scoped styles, we just went with the quick fix and scoped them here too.

Eventually, when https://homeday.atlassian.net/browse/FET-247 is done, we won't need two components.

This pull request does not affect the component visually.


aym3nb (approved):
Good catch 👀 

I QA'ed this change in myHomeday and everything is working as expected 👍 

Thanks for taking care of it.

PR #860, opened 2021-11-02
✨ Phone input component

SusCasasola:
This PR contains a new input component special for phone numbers. The purpose is to force users to use the dial code of their country. In the end is going to emit the value with the format we use on our databases for phone numbers.

The definition and other discussions are here: https://homeday.atlassian.net/browse/DS-202?atlOrigin=eyJpIjoiMjNkMGYyZWYxMDRiNDg2MmFjZmQ3MmU0Yzc1NzFlZGYiLCJwIjoiaiJ9

- The phone input component can receive a list of preferred countries to render them at the top. 
- It is supposed to be navigable with Tabs, Arrow keys and ESC key.

Let me know what can we do different or if something important is missing :)


ilyasmez:
Hey @SusCasasola, the component looks great! That should definitely increase the quality of the phone numbers that the users type.

I have a few questions:
- I see that you have the list of the countries hardcoded in the project (name in German, emoji, code...):
  - Did you consider using a 3rd party library for this, so we don't have to maintain this non-core data?
    Maybe we can use a combination of these two modules: https://github.com/michaelwittig/node-i18n-iso-countries/blob/master/langs/de.json and https://github.com/thekelvinliu/country-code-emoji
  - What about languages other than DE?
    The library I shared has many languages, so we can maybe the language dynamic. Or ask people to pass the list of countries in the language they want as a prop (to avoid a huge bundle size).
  - As far as I remember flag emojis are not supported on Windows, and I see that the input depends on them. Did you check the support of the flag emojis? We can also use SVGs instead (we can find a module for that).

- I see that you're overwriting some input stylings and using `!important` in some cases, what do you think of adding a prop to the component (e.g. `borderless` or `noBorderStyles`) to make it easy for you to style it without heavily depending on the classnames or using `!important`.

- And a very generic question (I think it's a bit too late to ask this one 😄), did we consider using a third party component instead of building our own? To avoid all the issues I mentioned above and to also have a proven accessibility patterns (like being able to search for the country by name instead of scrolling through the whole list).

SusCasasola:
Hey @ilyasmez 

Thank you for your review and your questions.

Regarding the hardcoded countries, I think that using a third party library for the list makes total sense, but it would need to include the dial code for each country, as well as the flags (using svgs is a better option, thanks for that!). By the way, I think that we need to make sure that we are going to need support for all countries, because I heard some comments saying that we are supporting only a few ones. If that’s the case, maybe we wouldn’t need to add a whole list here, the user could pass an array with the countries list as a prop (in the preferred language also).

Regarding the usage of a prebuilt component: To be honest, I didn’t consider using a third party component because I thought that it wasn’t allowed (I felt like it was cheating 😅), specially if I would need to force some styles to make it look like our inputs and interactions. I am going to take care of that feature with the keys though as I am already listening to the keyboard to handle the navigation with arrows.

ilyasmez:
Hi @SusCasasola :)

> Regarding the hardcoded countries, I think that using a third party library for the list makes total sense, but it would need to include the dial code for each country, as well as the flags (using svgs is a better option, thanks for that!).

Yes, and we can find a library that provides that (e.g. [what-country](https://www.npmjs.com/package/what-country)), I've mentioned `node-i18n-iso-countries` just for the translations.

> By the way, I think that we need to make sure that we are going to need support for all countries, because I heard some comments saying that we are supporting only a few ones.

That's the good thing about a third party library, we don't have to maintain just a political content 😄


> If that’s the case, maybe we wouldn’t need to add a whole list here, the user could pass an array with the countries list as a prop (in the preferred language also).

I veeery much agree with this idea!
I would also suggest to export a list of countries and their data directly in homeday-blocks, so the users can pass it as a prop if they want, but optionally. This way we ensure that we have a consistent list of countries in our products (homeday-blocks will be the source of this data), but also allow people to use their own list without having the whole list included in their bundle (if we export the data file separately, it will be tree-shakable).

What do you think @homeday-de/frontend-engineering?

> Regarding the usage of a prebuilt component: To be honest, I didn’t consider using a third party component because I thought that it wasn’t allowed (I felt like it was cheating 😅), specially if I would need to force some styles to make it look like our inputs and interactions. I am going to take care of that feature with the keys though as I am already listening to the keyboard to handle the navigation with arrows.

It's not cheating at all 😄 especially that handing phone numbers is not a core feature of the project. And there are few open source components that support a reasonable level of customization.
But we already have our own component now 🎉 we can forward with it and see how things go.

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,368 @@
+<template>
+  <div class="phone-input">
+    <button
+      class="phone-input__selector"
+      aria-haspopup="listbox"
```
volcanioo, line 5: You use `aria`!!! WE CARE ABOUT ACCESSIBILITY! 🎉 🎉 🎉 🎉 🎉  

YOU THE MVP!!!
MVP

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,368 @@
...
+      type="tel"
+      name="phoneNumber"
+      label="Telefonnumer"
+      v-model="phoneNumber"
+      class="phone-input__field"
+      placeholder="+55 555 555 5555"
+      :custom-rules="[rule]"
+      :formatter="phoneFormatter"
+      @input="handleInputEvent"
+    />
```
volcanioo, line 58: Why do we use `HdInputFormatter` component even we format our phone number via `phoneFormatter` function (::210)?

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,368 @@
...
+      <span :class="`phone-input__selector__flag flag-icon flag-icon-${selectedCountry.code.toLowerCase()}`"></span>
+      <HdIcon
+        width="24px"
+        height="24px"
+        :src="smallArrowIcon"
+        :class="arrowClassNames"
+      />
+    </button>
+    <ul
+      id="country-codes-list"
```
volcanioo, line 19: Do we need this `id` attribute?

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,368 @@
...
+      showDropdown: false,
+      phoneNumber: COUNTRY_PHONE_CODES.find((country) => country.code === this.defaultCountry)?.dial_code,
+      selectedCountry: COUNTRY_PHONE_CODES.find((country) => country.code === this.defaultCountry),
+      focusedCountry: null,
+      rule: {
+        validate: (value) => {
+          const hasMoreThan8Digits = value.length >= 8;
+          const hasPlusDigit = value.includes('+');
+          const isValid = hasPlusDigit && hasMoreThan8Digits;
+          return isValid;
```
volcanioo, line 134: Wow 🤓  I got inspired by your approach to managing the if condition here. Make total sense!

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,368 @@
...
+  },
+  methods: {
+    toggleDropdown() {
+      this.showDropdown = !this.showDropdown;
+    },
+    selectCountry(country) {
+      this.selectedCountry = country;
+      this.toggleDropdown();
+      if (this.phoneNumber.length <= MAX_DIGITS_DIAL_CODE) {
+        this.$refs.input.$refs.input.$refs.input.focus();
```
volcanioo, line 204: Isn't this reference too deep? Are we sure there will always be three-dimensional input?

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,368 @@
...
+    phoneFormatter(phone) {
+      return formatPhoneNumber(this.selectedCountry.dial_code, phone);
+    },
+    keysHandler(event) {
+      const { keyCode } = event;
+      const firstCountryCode = this.sortedCountriesList[0].code;
+      const lastCountryCode = this.sortedCountriesList[this.sortedCountriesList.length - 1].code;
+      let countryNode = null;
+
+      const foucsCountryCode = () => {
```
volcanioo, line 216: `foucsCountryCode` => `focusCountryCode` 🤓  - I am using cSpell on my vscode. I will share the URL with you in case if you want to see it as well: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,368 @@
...
+$dropdown-button-width: 74px;
+$dropdown-button-height: 55px;
+
+.phone-input {
+  position: relative;
+  display: flex;
+
+  ::v-deep {
+    .field__main .input {
+      border-top-left-radius: 0 !important;
```
volcanioo, line 273: every-time-you-use-important-in-css-1-kitten-dies

I just wanted to make sure that you are informed about this situation. 😄  Avoiding the usage of `!important` is essential, but it is also okay if you don't have any other way to go. :)

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,368 @@
...
+    left: 0;
+    z-index: 10;
+    width: 100%;
+    max-height: 264px;
+    overflow: auto;
+    border-radius: $sp-xs;
+    background-color: $white;
+    box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.5);
+    &__option {
+      .option-country {
```
volcanioo, line 333: Current BEM Structure:

- `phone-input__dropdown`: `<ul>`
    - `phone-input__dropdown__option`: `<li>`
        - `option-country`: `<button>`
            - `option-country_name`: `<span>`, `<p>`, etc.

What do you think about having something like this instead? ⏬ 

- `phone-input__dropdown`: `<ul>` ✅ 
    - `phone-input__dropdown__option`: `<li>` ✅ 
        - `option__country`: `<button>` 🟧 
            - `option__country__name`: `<span>`, `<p>`, etc.  🟧

Furthermore, you can check this exact section of the article to also get more information regarding this topic: https://www.smashingmagazine.com/2016/06/battling-bem-extended-edition-common-problems-and-how-to-avoid-them/#8-how-to-nest-components

src/services/utils.js
```diff
@@ -211,7 +211,35 @@ export function keysToSnakeCase(collection) {
...
+    const groupsOfNumbers = phoneNumber.substring(countryCodeDigits).match(/(\d{0,3})(\d{0,3})/);
+    return `${countryCode} ${groupsOfNumbers[1]} ${groupsOfNumbers[2]}`;
+  } else if (phoneNumber.length > countryCodeDigits + 6 && phoneNumber.length <= countryCodeDigits + 10) {
+    const groupsOfNumbers = phoneNumber.substring(countryCodeDigits).match(/(\d{0,3})(\d{0,3})(\d{0,4})/);
+    return `${countryCode} ${groupsOfNumbers[1]} ${groupsOfNumbers[2]} ${groupsOfNumbers[3]}`;
+  } else {
+    const groupsOfNumbers = phoneNumber.substring(countryCodeDigits).match(/(\d{0,3})(\d{0,3})(\d{0,4})(\d{0,10})/);
+    return `${countryCode} ${groupsOfNumbers[1]} ${groupsOfNumbers[2]} ${groupsOfNumbers[3]}${groupsOfNumbers[4]}`;
+  }
+}
```
volcanioo, line 240: It might become very tricky to develop a formatted function from the stretch as it has many decisions that need to be discussed. What about using a **tiny** 3rd party library to manage formatting a phone number? For example, https://github.com/grantila/awesome-phonenumber#readme (I didn't check if it is the best option)

volcanioo (commented):
hacking

This component will be used in many places across our products. Thank you so much for your great effort!! 🎉 🎉 🎉   In addition, I would like to point out a couple of things here and ask further questions to understand your approach in a better way.

Before we pass the questions, I should say, I agree with all comments of @ilyasmez. He talks good stuff here as usual! 

* I agree with @ilyasmez 's comments regarding using 3rd party library instead of having a hardcoded approach;
    * to increase the code readability
    * to not to spend lots of time on compatibility topics such as support on different browsers and screen sizes
    * to stay up to date without wasting too much developer time (this kind of small 3rd party library will not require us to do a refactoring on the phone input component after two years, for instance: https://www.npmjs.com/package/vue-tel-input)

and also, especially, `!important` topic! :) 

<hr/>

Another opportunity that I feel like we can remove lots of code lines by using a 3rd party library from this component.  

<hr/>

I asked multiple questions, don't hesitate to let me know if something does not make sense to you. We can discuss, and I can share more resources with you :) Thank you so much again for your excellent work! You rock!!! 🚀 🚀 🚀 🚀

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,368 @@
+<template>
+  <div class="phone-input">
+    <button
+      class="phone-input__selector"
+      aria-haspopup="listbox"
+      aria-label="Choose a country code:"
```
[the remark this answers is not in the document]
SusCasasola: I agree totally!

src/components/form/HdInputPhone.vue
SusCasasola: Oh, the HdInputFormatter needs a function to format the value as we want, that's why it's there

src/components/form/HdInputPhone.vue
SusCasasola: 😅 thanks!

src/components/form/HdInputPhone.vue
SusCasasola: Oh yes, I actually was going to ask if there is a better way to handle this, I feel that it's super insecure doing this, but so far yes, we always are going to have three inputs.

src/components/form/HdInputPhone.vue
SusCasasola: Yes! Thank you 🙏🏼

src/components/form/HdInputPhone.vue
SusCasasola: I know I hate using this one haha I am going to see if I have a workaround, I think that adding more specificity to the selector should work 👍🏼

src/components/form/HdInputPhone.vue
SusCasasola: oh nice! I agree, I will change it

src/services/utils.js
SusCasasola: Totally agree! specially now that we are going to add the zero for the format

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,368 @@
...
+      @click="toggleDropdown"
+      :aria-activedescendant="`${selectedCountry.dial_code}, ${t.COUNTRIES[selectedCountry.code]}`"
```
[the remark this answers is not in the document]
SusCasasola: For sure!

src/components/form/HdInputPhone.vue
SusCasasola: Actually I put it there while I was adding the aria labels, but I was changing the structure while I was coding, so in the end I think that I forgot why it was there at the beginning 😅 I think I will just remove it

src/components/form/HdInputPhone.vue
volcanioo: Hmm, interesting, I don't know if we need `HdInputFormatter` component in this case, but don't worry, each solution works for me, I will resolve the conversation, thanks for your answer.

src/components/form/HdInputPhone.vue
volcanioo: hmm.. my idea would be; we can get rid of the `HdInputFormatter` component in this case. I can't see a necessity to use it as we already define the formatting function in the `HdInputPhone.vue` - so, we can remove the `HDInputFormatter` and use the formatting function to format the phone without passing the function through another component. in that way, we can get rid of nested inputs as well. WDYT? does it make sense?

src/services/utils.js
SusCasasola: Actually the library is really good! I used it in other sections where I needed to validate the dial code, thanks for this suggestion 🥳

src/components/form/HdInputPhone.vue
ilyasmez: I think the usage of `HdInputFormatter` is fine here, or maybe even recommended. It's a simple component that does a simple and specific job. Removing it will require us to replicate its functionality in HdInputPhone, which goes against reusability and modularity.

I think the deep access to the grandchildren is okay here.
A better option would be to expose a method `focusInput` from HdInputFormatter and HdInput, so you can do just `this.$refs.input.focusInput()`, but that's still relaying on the internal API of the component which is not recommended, as it might change in the future without any notice.
An even better option would be to use the exposed API of the component, aka the contract (props and events): We can have a `focused` prop (might even use the native `autofocus`, but I didn't test it), with `false` as the default value, and we can dynamically set it to `true` when we want to focus the component, and of course always set it back to `false` when the input has been blurred.

ilyasmez:
Hi @SusCasasola 🙂 What's the status of this PR? Is it ready for another review?

src/components/form/HdInputPhone.vue
SusCasasola: I can try setting the `focus` prop and see how it works, thank you @ilyasmez!
And yes @volcanioo, the reason I used the `HdInputFormatter` in the beginning was to avoid doing the same logic of format the value only when it's not focused

SusCasasola:
Hey @ilyasmez! yes, it's ready for another review

volcanioo (approved):
Thanks for your effort! :) I just asked a couple of questions and put optional change requests. Please feel free to merge your ticket if you do not have time for them :) 

ok-lets-do-it

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,383 @@
...
+      </template>
+    </ul>
+    <HdInputFormatter
+      ref="input"
+      type="tel"
+      name="phoneNumber"
+      label="Telefonnumer"
+      v-model="phoneNumber"
+      class="phone-input__field"
+      placeholder="+55 555 555 5555"
```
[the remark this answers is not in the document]
SusCasasola: It might be confusing if this input would be used only for German numbers, but the intention is to use international numbers too, that's why we used a generic placeholder... but now that you mention it, I can use the selected dial code to modify the placeholder too, and it will be updated everytime we select another one 👍🏼

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,383 @@
...
+      showDropdown: false,
+      phoneNumber: COUNTRY_PHONE_CODES.find((country) => country.code === this.defaultCountry)?.dial_code,
+      selectedCountry: COUNTRY_PHONE_CODES.find((country) => country.code === this.defaultCountry),
+      focusedCountry: null,
+      rule: {
+        validate: (value) => {
+          const phoneNumber = new PhoneNumber(value);
+          return phoneNumber.isValid() && phoneNumber.isMobile() && phoneNumber.canBeInternationallyDialled();
+        },
+        errorMessage: 'e.g. +55 555 555 5555',
```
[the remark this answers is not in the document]
SusCasasola: I agree! Let me update these

src/components/form/HdInputPhone.vue
```diff
@@ -0,0 +1,383 @@
...
+      return [baseClass, directionClass];
+    },
+    selectedFlagClassNames() {
+      const baseClass = 'phone-input__selector__flag';
+      const countryClass = `flag-icon flag-icon-${this.selectedCountry.code.toLowerCase()}`;
+      return [baseClass, countryClass];
+    },
+    activeCountryCodeAndName() {
+      return `${this.selectedCountry.dial_code}, ${this.t.COUNTRIES[this.selectedCountry.code]}`;
+    },
```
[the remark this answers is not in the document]
SusCasasola: Oh I could have used a third party library for all the feature but at the beginning I didn't consider it because I felt that I was cheating 🙈 I didn't know that we were allowed to do that because it would make the library bigger (due to the dependencies), and also I thought that it was better if we could configure certain things on our own if those were too specific. I also wanted to use our own Input components as it already has a lot of logic for formatting and passing value, labels and other props.

kuroski (approved):
It's a lot of work to develop a phone input, so thank you so much for this great addition! 😄 

I have a few comments on the code, let me know your opinion =D

I have also some extra observations, they can serve as `"next interactions"` for this component.

Please let me know your opinion and if you think they are relevant 😄 

### Keyboard navigation

- The country list keyboard navigation seems a bit off, if I keep pressing the down key in the list, it will keep scrolling, but the selection will just get the next item
  - The scroll is also faster than the selection, check the video bellow

https://user-images.githubusercontent.com/1855125/150753280-3274f60c-9865-4309-9f56-dc4debeac9cb.mov

> Here I'm just pressing the "down key".

### "Performance"

- We can invest in using virtual scrolling and caching for the country list

Check the movie below:

https://user-images.githubusercontent.com/1855125/150753923-1ce15bff-dadb-469a-8fc3-34233db23382.mov

We are loading ~431kb of svg's with more than 200 request when we open the list for the first time:

image

Aside from that, the list is fully loaded (despite only showing just a part of it).

That can cause problems to users using slower mobile devices and that are within slower internet connections.

We, later on, can think about using virtual scrolling and maybe working on a caching mechanism that would help us load that.

---

Thanks again for that!
Feel free to reach me anytime.
I will also approve the PR since I don't feel that my comments should be blockers.

🚢

SusCasasola:
@kuroski Thank you so much for your review! 🙏🏼 I will create a Jira card to manage all this improvements for another PR, I will post it here when it's ready

PR #866, opened 2021-11-08
🐛  Avoid overglow of HdToggle icon

drapisarda:
As you can see, the chevron icon can overflow because of sizing. This avoids effecting the outher layout.

image

### DIFF
--- a/src/components/HdToggle.vue
+++ b/src/components/HdToggle.vue
@@ -212,7 +212,7 @@ $_controlIconSize: 32px;
     text-align: left;
     transition: outline $time-s ease-in-out;
 
-    // avoids to expand out of the content due to animation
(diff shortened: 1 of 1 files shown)


leandroinacio (approved):

ilyasmez:
Hey @drapisarda, I have one question, how did you end up with the icon occupying the full height? From the code, I see that it has a fixed width and height (`$_controlIconSize`). And it's also absolutely positioned, so it's always on top, while in your example it's vertically centered. Did you override the styles in your local project?

I just want to understand the problem better, to see if the `overflow: hidden` is the way to go.

drapisarda:
@ilyasmez thank you for checking. I checked and actually is how I vertically centered than makes it bad. Therefore this can be closed. Thank you for your support.

PR #868, opened 2021-11-12
♿️ Add type button for Calendar, Modal, Toast and ArrowButton components

kuroski:
Buttons should have an explicit `type` attribute defined, otherwise, if we add this component inside of a form, the button will automatically behave like a submit button.

This PR is to fix a problem with our current task: 
> we are rendering the `HdCalendar` inside of a form, but when we click on the "arrow button" inside of the calendar, it is triggering a form submission.

We took advantage and also added a `type="button"` for other general components.

### DIFF
--- a/src/components/HdCalendar.vue
+++ b/src/components/HdCalendar.vue
@@ -2,6 +2,7 @@
   <div class="calendar">
     <header class="calendar__header">
       <button
+        type="button"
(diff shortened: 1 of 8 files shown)


ilyasmez (approved):
💯

petterHD (approved):
🚢

PR #871, opened 2021-11-17
⬆️ Upgrade dependencies

SusCasasola:
This PR is meant to upgrade some dependencies that can help with some of the active dependabot alerts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "@storybook/addon-actions": "6.2.9",
     "@storybook/addon-console": "1.2.3",
     "@storybook/addon-essentials": "6.2.9",
-    "@storybook/addon-knobs": "6.2.9",
+    "@storybook/addon-knobs": "6.3.1",
     "@storybook/addon-links": "6.2.9",
     "@storybook/addon-viewport": "6.2.9",
     "@storybook/theming": "6.2.9",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
     "@storybook/addon-actions": "6.2.9",
     "@storybook/addon-console": "1.2.3",
     "@storybook/addon-essentials": "6.2.9",
-    "@storybook/addon-knobs": "6.2.9",
+    "@storybook/addon-knobs": "6.3.1",
```
viniciuskneves, line 52: Hey @SusCasasola I would recommend you to check https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-62x-to-630

It might be worth going for Storybook 6.3 instead as it deprecates some addons and I'm not fully sure about compatibility once it comes to Storybook minor upgrades.

Let me know what you think!

package.json
```diff
@@ -58,8 +58,8 @@
     "@types/lodash": "^4.14.171",
     "@vue/cli-plugin-babel": "~4.5.13",
     "@vue/cli-plugin-eslint": "~4.5.13",
-    "@vue/cli-plugin-unit-jest": "~4.5.13",
-    "@vue/cli-service": "~4.5.13",
+    "@vue/cli-plugin-unit-jest": "^4.5.15",
+    "@vue/cli-service": "^4.5.15",
```
viniciuskneves, line 62: 👌

package.json
SusCasasola: Thanks a lot @viniciuskneves! You are right, it's seems that they are replacing the knobs with controls, so I will need to update the components that are using knobs.

PR #878, opened 2021-12-09
✨ Add TypeScript support

aym3nb:
## Why
Currently, TypeScript is supported in the [Website](https://github.com/homeday-de/website), [Customer App](https://github.com/homeday-de/customer-app), and some other projects. 
And we are already using JSDoc to do some type checking.
So it made sense to add its support here to be able to benefit from its capabilities.

---

## Changes
- Add TypeScript support:
  - Add `Vue-CLI` TypeScript plugin
  - Remove `jsconfig.json`
  - Change `target` to `es2019` to fix the issue: https://tinyurl.com/ts-additional-loader-issue
  - Configure TS parser in Eslint
  - Disable a set of Eslint rules that TS compiler already validates
  - Disable a few strict rules to allow people to use TS in a less strict mode
  - Fix Eslint throwing import/no-unresolved error on declaration files
  - Extend main Eslint config in jest's Eslint config
- Convert `main.js` to TS
- Convert `src/lang/index.js` to TS
- Create a declaration file for global types `src/@types/global.d.ts` & Add `customRules` type in it
- Use TS instead of JSDoc in `HdRadioCard` component, convert its test to TS , and update its snapshot
- Export `HdRadioCard`component name to be able to use it in `HdRadioCardGroup`
- Use TS instead of JSDoc in `HdRadioCardGroup` component, convert its test to TS , and update its snapshot
- Use TS instead of JSDoc in `HdCheckboxCard` component, convert its test to TS , and update its snapshot
- Export `HdCheckboxCard`component name to be able to use it in `HdCheckboxCardGroup`
- Use TS instead of JSDoc in `HdCheckboxCardGroup` component, convert its test to TS , and update its snapshot
- Remove unnecessary `@ts-check` in `HdCheckboxIndicator`
- Add a new script `build:tsc` to compile TS to JS
- Add a step to compile TS before release and publish
- Fix `build:lib` script, some lint errors ,and some failing tests due to the error can't find homeday-blocks in X
- Add basic TypeScript [documentation](https://github.com/homeday-de/homeday-blocks/blob/add-typescript-support/docs/TypeScript.md#typescript-ts)


ivandata (approved):
WOW! Amazing changes! All looks good!

QQ: Is it possible to avoid using // @ts-ignore in the main.ts file?

aym3nb:
> QQ: Is it possible to avoid using // @ts-ignore in the main.ts file?

Without ignoring those lines, the TS compiler would show this error in VSCode:
`Ignore TS on the imports that cause the problem: Module '"*.vue"' has no exported member 'TYPES'. Did you mean to use 'import TYPES from "*.vue"' instead? ts(2614)`

 I tried a couple of methods but none solved it, did you face such error before?

ivandata:
> > QQ: Is it possible to avoid using // @ts-ignore in the main.ts file?
> 
> Without ignoring those lines, the TS compiler would show this error in VSCode:
> 
> `Module '"*.vue"' has no exported member 'TYPES'. Did you mean to use 'import TYPES from "*.vue"' instead? ts(2614)`
> 
> I tried a couple of methods but none solved it, did you face such error before?

In my IDE I didn't get the error but I think it can be solved by importing modules first before they will be exported. 
Something like:
```
import HdLink, { TYPES as HdLinkTypes } from './src/components/HdLink.vue';
export { HdLink, HdLinkTypes };
```

I know it's can be out of the scope of this ticket, but in my opinion, in addition to removing `ts-ignore`, it also makes the main file more readable.

ivandata:
Forgot to ask, what about backward compatibility? As far as I know, most of the modules on projects use js import. 
The release version depends on the answer according to https://semver.org/

aym3nb:
> In my IDE I didn't get the error but I think it can be solved by importing modules first before they will be exported. 

Unfortunately, the error remains but since it happens only in my IDE, I removed the comments 1093c96

> Forgot to ask, what about backward compatibility? 

I think the new version would be minor, I tested this version in customer-app it didn't require any changes.

ivandata:
> > In my IDE I didn't get the error but I think it can be solved by importing modules first before they will be exported.
> 
> Unfortunately, the error remains but since it happens only in my IDE, I removed the comments [1093c96](https://github.com/homeday-de/homeday-blocks/commit/1093c96e25eb97db486670f1fb64ab65abdc043f)
> 
> > Forgot to ask, what about backward compatibility?
> 
> I think the new version would be minor, I tested this version in customer-app it didn't require any changes.

We need another reviewer, bc I'm using JetBrains. But it can be VS code issue.

ilyasmez (commented):
That's a huuuge step towards a better developer experience, and consequently a better user experience. Kudos to you @aym3nb! 🎉 

I have just one comment:
I failed to find in the changes a place where you compile the `ts` files.
Because without it the library will work only for the projects that support TS already, and that will exclude the other projects where we're not supporting TS yet at Homeday or any external project.
What I would suggest would be to add a step to compile the project [before creating a release](https://github.com/homeday-de/homeday-blocks/blob/master/.github/workflows/release.yml#L29) and publishing it on NPM. This will also require setting the `main` file in the package.json back to `main.js`.

aym3nb:
Thank you so much for the review @ilyasmez 

> I failed to find in the changes a place where you compile the `ts` files.
> Because without it the library will work only for the projects that support TS already, and that will exclude the other projects where we're not supporting TS yet at Homeday or any external project.
> What I would suggest would be to add a step to compile the project [before creating a release](https://github.com/homeday-de/homeday-blocks/blob/master/.github/workflows/release.yml#L29) and publishing it on NPM. This will also require setting the `main` file in the package.json back to `main.js`.

Great catch, I totally overlooked that 😅 

I made the suggested changes 👍

ilyasmez:
> Thank you so much for the review @ilyasmez
> 
> > I failed to find in the changes a place where you compile the `ts` files.
> > Because without it the library will work only for the projects that support TS already, and that will exclude the other projects where we're not supporting TS yet at Homeday or any external project.
> > What I would suggest would be to add a step to compile the project [before creating a release](https://github.com/homeday-de/homeday-blocks/blob/master/.github/workflows/release.yml#L29) and publishing it on NPM. This will also require setting the `main` file in the package.json back to `main.js`.
> 
> Great catch, I totally overlooked that 😅
> 
> I made the suggested changes 👍

Great! Thank you @aym3nb!

I still think it won't work properly yet, because we have `"outDir": "./dist"` in tsconfig, which means that the compiled files will be moved to the `dist` directory, and this is not something we want, because we want the structure of the project to stay the same, as all the imports depend on it.
I can think of two solutions:
- Option 1: We compile the files in place (so the `outDir` can be `./` instead), this way we keep the same structure for the project. But I guess this will cause some errors because you'll be overwriting the source files, which is a bit tricky, but it won't harm as we do it just on the CI level.
- Option 2: Keep the `outDir` as `./dist`, and merge the compiled and uncompiled files in `./dist` (the js files, package.json...), and then publish the content of `./dist` to npm, instead of the project's root directory.

I would also suggest to debug the solution in a new Vue project that does have typescript support yet.

Let me know what do you think.

aym3nb:
> Let me know what do you think.

Thanks again @ilyasmez  🙇  I agree with you, and I decided to go with the second option.

I adjusted the build scripts to:
```json
    "prebuild": "del dist/",
    "build": "tsc",
    "postbuild": "cpy '**/*' '!**/*.{ts,tsx,js,jsx}' ../dist/src/ --cwd=src/ --no-overwrite --parents",
```

And added another step in the release GH workflow to [find and replace](https://github.com/marketplace/actions/find-and-replace) `main` in `package.json` 

---

> I would also suggest to debug the solution in a new Vue project that does have typescript support yet.

And while debugging in a new Vue project, it seems like there is another issue with Vue components written in TS like [HdRadioCard](https://github.com/homeday-de/homeday-blocks/blob/add-typescript-support/src/components/form/HdRadioCard.vue#L49) because they are not compiled with TS compiler so they cause a parsing error.

Any suggestions on what can be done to solve this? or should client projects support TS in order to consume blocks from now on?

ilyasmez:
> And added another step in the release GH workflow to [find and replace](https://github.com/marketplace/actions/find-and-replace) `main` in `package.json`

You can just set this field to `./dist/main.js` directly in package.json, no need to do it right before the build. Because this field (`main`) is useful just when the package is used externally.

> And while debugging in a new Vue project, it seems like there is another issue with Vue components written in TS like [HdRadioCard](https://github.com/homeday-de/homeday-blocks/blob/add-typescript-support/src/components/form/HdRadioCard.vue#L49) because they are not compiled with TS compiler so they cause a parsing error.
> 
> Any suggestions on what can be done to solve this?

Oh yeah that's true, because TS doesn't know about the Vue syntax, it's vue/ts loader that takes care of that. So `tsc` wouldn't result in .vue files written in JS.
I found this project [vue-compile](https://github.com/egoist/vue-compile) which seems to do what we need in this case. You might want to give it a try.

If not then we can think of compiling the library differently.

> or should client projects support TS in order to consume blocks from now on?

Adding TS as a requirement to use homeday-blocks would introduce a bad DX imho. And we will depend on the tsconfig of the consuming project, so we might require some specific rules to be added there, and that might result in so much friction to start using the library :/ I'd prefer if we take care of the transpilation from our side.

kuroski:
Heello, I'm back from my vacation, and maybe I can help out with this topic 😄 

First of all, thaaanks for the great initiative @aym3nb!!

I will try to answer the things I've read at once, check the topics bellow

### Removal of `@ts-ignore`s

Regarding the `@ts-ignore`, it "makes sense to use them in this case" (notice the quotes here). 

Vue files currently do not support type exports, maybe Jetbrains IDEs deal with that in a more intelligent way, but VScode is a bit "less-complete" in this scenario since is just a text editor on steroid, this is why `@ts-ignore` is required to make it work and why the suggesting made on "importing-first then exporting" won't work.

More on that [here](https://github.com/vuejs/vue-loader/issues/1281).

To actually remove the `@ts-ignores` in this case we could generate declaration files for vue files, or enhance the shims files by hand (which is not the best way and I think it wouldn't worth it). 

Probably the best way would be extracting those "TYPES" into its own TS files.

### Building Homeday blocks

As @ilyasmez mentioned, we would have to build Homeday blocks in order to make it work properly.

I would recommend to compile everything, there are a few things we could finish up... @aym3nb, we could schedule some day during this month to pair, and maybe we can finish up the build script together, what do you think? I can help out with that 😄

aym3nb:
Welcome back 👋  and thank you so much for the review @kuroski 

> Probably the best way would be to extract those "TYPES" into its own TS files.

Great, I'll extract them then 👍 

> I think there are a few things that we would need to do in order to finish up... @aym3nb, we could schedule someday during this month to pair, and maybe we can finish up the build script together, what do you think?

Yes, of course, that would be great 👍

aym3nb:
Hey, @homeday-de/frontend-engineering this PR is ready to be merged since everything is working as expected and all concerns were addressed.

I’d appreciate it if you could take some time to review the PR after the last changes 🙇

leandroinacio (approved):
Another amazing pull request Aymen! 👏 
🚀

PR #892, opened 2022-01-11
💥 Use Vue-CLI to build the library

aym3nb:
### Intro

After adding [TS support](https://github.com/pulls/878) to the project our build process became longer and more complicated (type-checking, compiling, transpiling), we found that we would need some additional tools (e.g. [Esbuild](https://esbuild.github.io/)) to successfully build the library but the issue is most tools don't support Vue at all or just Vue 3.

On the other hand, we could use Vue-CLI which can do all the necessary steps to build the library without any additional tools.

If we use Vue-CLI to build the library with our entry file `main.js` it would bundle the whole library in one big file therefore no tree-shaking.

So to avoid sacrificing tree-shaking, Ilyas suggested building each component separately then generating a file that exports them.

---

###  Changes
- Fixed some linter errors
- Extracted component types into separate files like what was done in the TS [PR](https://github.com/homeday-de/homeday-blocks/pull/878/commits/12224a5cd6a053f52b218ef33f0bbf116a3f255b)
- Added a script that builds all the library files separately using Vue-CLI then generate the main.js file
  The execution time is around **1 ~ 1,5 minutes** on MacBook Pro (2019) | 2.8 GHz Quad-Core Intel Core i7 | 16 GB RAM
- Added a step to build the library before releasing it in the release workflow
- Deleted main.js file since it is now auto-generated by the build script
- Updated the [documentation](https://github.com/homeday-de/homeday-blocks/blob/390cb1dd66daf610516ff703fc253327614190cd/docs/BUILD.md)
- Updated Hygen scaffolding templates

---

###  Breaking Changes for consuming projects
- Components and services relative import paths won't work anymore, e.g.
`import { DateService } from 'homeday-blocks/src/services/date'` should be `import { DateService } from 'homeday-blocks'` instead.

---

Screen Shot 2022-02-01 at 08 23 15


ivandata:
If it works fast and allows us to use tree-shaking with TS I'm all for it. 
In any case, it would be very interesting to compare benchmarks using Vue-CLI and Webpack + Rollup approach.

leandroinacio:
Hey @aym3nb this looks amazing ❤️
Do you plan on adding some documentation about how this new build flow works?

aym3nb:
@leandroinacio Thank you so much for the review

> Do you plan on adding some documentation about how this new build flow works?

So far, I only made sure to document the whole script. but sure I can also add more information about it in the documentation.

leandroinacio:
@aym3nb given that github actions will take care of building and publishing, shouldn't we add `main.js` to `.gitignore` to avoid merging the file?

aym3nb:
> given that github actions will take care of building and publishing, shouldn't we add `main.js` to `.gitignore` to avoid merging the file?

Yes good catch, I'll add it to `.gitignore`

leandroinacio:
@aym3nb sorry about the delay to give approval.
I did some tests on sliders and the website projects. 
Sliders works fine, and the bundle size doesn't change much. The website on the other hand, I get an error.

```
error git dep preparation failed
```

I've already tried clearing cache and other stackoverflow solutions. The only thing that works is downgrading node to version  v14.17.5 (we use v16.12.0 in the website). Any ideas on what could be the cause?

aym3nb:
Thanks, @leandroinacio yeah, I also tested only myHomeday and another testing project but none of them use node higher than v14.

> Any ideas on what could be the cause?

Not really, maybe someone from @homeday-de/frontend-engineering does.

ivandata:
> Any ideas on what could be the cause?

Sorry, but the cause may be the version of the node, and that's normal. 😅

ivandata (approved):
Awesome job, Aymen! Looks great to me. 🥇

leandroinacio (approved):
As pointed out by Aymen here https://github.com/homeday-de/website/pull/869#issue-1141344791 , issue resolved.
🚀

PR #917, opened 2022-02-24
🐛 Fix flag-icons-svg styles import paths in HdInputPhone

aym3nb:
In this PR [HdInputPhone ](https://github.com/homeday-de/homeday-blocks/blob/master/src/components/form/HdInputPhone.vue)component `flag-icons-svg` styles were fixed because they were being imported with relative paths that can't be resolved when using homeday-blocks in other projects.


leandroinacio (approved):

limhenry (approved):
Issue fixed! (Missing `flag-icons-svg` module). 🎉 Thank you!

PR #929, opened 2022-03-09
⚡️ Limit build script concurrency 

aym3nb:
**Why limit build script concurrency?**
Because it was too much for the Github-hosted runners which are used to run actions (2-core CPU / 7 GB RAM)

### Changes
- Added [p-limit](https://www.npmjs.com/package/p-limit) as a dev dependency
- Used `p-limit` to limit the build script concurrency based on the number of CPU cores

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1959,6 +1959,17 @@
           "dev": true,
           "requires": {
             "p-limit": "^2.0.0"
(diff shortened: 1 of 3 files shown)


ivandata (approved):
LOVE IT! 🍔

PR #951, opened 2022-04-08
:goal_net: [CL-2994] TypeError: undefined is not an object (evaluating 'this.$refs.flickity.slides')

ivandata:
## Details
Fix the most popular error in the Sentry https://sentry.io/organizations/homeday-de/issues/?project=1381290&query=slides&statsPeriod=14d

## Related tickets
[CL-2994](https://homeday.atlassian.net/browse/CL-2994)
Sentry Issue: [MEINHOMEDAY-49R](https://sentry.io/organizations/homeday-de/issues/3066658413/?referrer=jira_integration)

## Main changes
- added condition for $refs at the mounted cycle


src/components/gallery/HdGalleryCarousel.vue
```diff
@@ -130,9 +130,11 @@ export default {
   },
   mounted() {
     this.$nextTick(() => {
-      this.updatePagerItemsCount();
-      this.$refs.flickity.on('change', this.updateCurrentIndex);
-      this.$refs.flickity.on('staticClick', this.onStaticClick);
+      if (this.$refs.flickity) {
```
volcanioo, line 133: https://stackoverflow.com/a/3390426/9218664
Screenshot 2022-04-11 at 12 32 24 PM


Shouldn't we check the type here? something like:
```if (typeof this.$refs.flickity !== 'undefined')```

**Review by volcanioo — APPROVED** [review]

File: src/components/gallery/HdGalleryCarousel.vue (hunk 1 above)
Comment (ivandata) [reply src/components/gallery/HdGalleryCarousel.vue:133]: In this case, we do not need to check the type of the object, we need to know if it is present at all. And there can be two options: either we check for the true or the false. I've chosen true checking and run the function only if the object is present. The following examples provides a complete list of JavaScript falsy values:
```
if (false)
if (null)
if (undefined)
if (0)
if (-0)
if (0n)
if (NaN)
if ("")
```
Therefore, the code will run the value is present, and will not be undefined. 😉

aym3nb (approved):
🚀

majakomel (approved):

PR #954, opened 2022-04-11
Remove scss scoped attrinutes

hesambayat:
- Removing some components style scoped attr that will snapshot check
- Adding boolean type to the HdRadio.value since some product are passing value as prop


aym3nb (approved):
👏🚀

PR #959, opened 2022-04-22
New Role Attribute for Checkbox Element

volcanioo:
## Details
We need to add `role` attribute to the actual input element to be able to reach the checkbox element via the `role` attribute in the customer app. I checked the lighthouse and couldn't see any issue with it, please let me know if you see anything.

## Related tickets
* [CL-2798](https://homeday.atlassian.net/browse/CL-2798)

I need this change to be able to query the checkbox element in our test suites of the customer app. We use the testing library and [we can only find elements by their roles](https://testing-library.com/docs/react-testing-library/cheatsheet/#queries). This is the reason why I need to duplicate the role attribute of the HdCheckbox element.

## Main changes
- added role attribute for the checkbox element

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7297,6 +7297,16 @@
           "integrity": "sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw==",
           "dev": true
         },
+        "ansi-styles": {
+          "version": "4.3.0",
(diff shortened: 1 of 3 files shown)


leandroinacio (approved):
If it doesn't affect our accessibility and SEO points in lighthouse, I'm fine with it 👍 

PS.: We can also use getByTestId (it will search for data-testid attribute, which you can add directly from customer-app level, HdCheckbox will inherit).

ivandata (approved):
These are good improvements. Using semantic attributes is better than a data attrs definitely. Thanks for taking care of it. LGTM 🏋️‍♂️

PR #965, opened 2022-04-26
✨ New Attr for Checkbox

volcanioo:
## Details
I just merged a change that adds a `role` attribute for the checkbox element. But, this solution didn't work in our side as expected. As a solution, I needed to replace the `role` attribute with a `data-testid` for the actual checkbox element. 

###### More info
I need this change to be able to query the checkbox element in our test suites of the customer app. https://testing-library.com/docs/queries/bytestid/

* #961 

## Related tickets
* [CL-2798](https://homeday.atlassian.net/browse/CL-2798)

## Main changes
- removed role attribute for the checkbox element
- added `data-testid` attribute for the element

### DIFF
--- a/src/components/form/HdCheckbox.vue
+++ b/src/components/form/HdCheckbox.vue
@@ -28,6 +28,7 @@
         class="checkbox__input"
         type="checkbox"
         :name="name"
+        role="checkbox"
         :disabled="disabled"
         v-model="isChecked"/>
       <div
--- a/src/components/form/HdCheckbox.vue
+++ b/src/components/form/HdCheckbox.vue
@@ -28,9 +28,10 @@
         class="checkbox__input"
         type="checkbox"
         :name="name"
-        role="checkbox"
         :disabled="disabled"
-        v-model="isChecked"/>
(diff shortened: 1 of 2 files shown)


ivandata:
Why in my opinion we don't need this attr. That doesn't mean the attribute is bad or something. It's just the last thing that we should use to get access to the DOM. Let me try to explain it.
First of all, as noticed by the link in the documentation: 
> In the spirit of [the guiding principles](https://testing-library.com/docs/guiding-principles), it is recommended to use this only after the other queries don't work for your use case.
Screenshot 2022-04-26 at 13 13 04


To get access to the dom elements recommended using `...Role` queries.
All queries can be founded [here](https://testing-library.com/docs/queries/about/). 

The second one, I don't understand the proposal for this. As I understand, this component uses a div element with all necessary attributes and events to work with form, so you don't need to get access to the checkbox element in most of all cases, it will be implementation details.

That doesn't mean the attribute is bad or something. It means I don't see any benefits in it. Maybe other devs will find it helpful.

PR #981, opened 2022-06-15
:green_heart: Add cleanup proccess to pr-deploy

leandroinacio:
**Ticket**
PEW

**Why?**
We have way too many objects deployed already, and no cleanup action.
So this will take care here, and one will be created to handle customer-app.
s3 console aws amazon com_s3_buckets_frontend-pulls_region=us-east-1 prefix=homeday-blocks_ showversions=false

**What?**
Cleanup github action. It will be executed when the pull request gets closed.

### DIFF
--- a/.github/workflows/post-pr-cleanup.yml
+++ b/.github/workflows/post-pr-cleanup.yml
@@ -1,0 +1,30 @@
+# This workflow takes care of cleaning up the pull request deploy for a clean S3 bucket.
+name: Post-PR cleanup
+
+on:
+  pull_request:
+    types: [closed]
+    branches:
+      - master
+
+env:
+  BUILD_DEST_DIR: ${{ github.event.repository.name }}/${{ github.event.pull_request.number }}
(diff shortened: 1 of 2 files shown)


.github/workflows/post-pr-cleanup.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:
+  pull_request:
+    types: [closed]
+    branches:
+      - master
+
+env:
+  BUILD_DEST_DIR: ${{ github.event.repository.name }}/${{ github.event.pull_request.number }}
+  S3_BUCKET: frontend-pulls
```
aym3nb, line 12: There is an existing secret for the bucket name.
```suggestion
  S3_BUCKET: ${{ secrets.AWS_S3_PR_DEPLOYMENT_BUCKET }}
```

aym3nb (approved):
Nicely done 👏 🚀

SusCasasola (approved):
Thanks for this change! 🚀

PR #987, opened 2022-06-29
:sparkles: Add support to dynamic forms for input formatter

hesambayat:
According to the conversation [here](https://homeday.atlassian.net/browse/VIEW-708?focusedCommentId=33539), we want to format user entries for area and price on search widget


aym3nb (approved):
🚀

PR #1007, opened 2022-07-26
✨  Add reference to hd details table

REX-Alexander:
Adds support for a reference to target separate instances of Hd Details Table. A requirement for: [#REAL-4538](https://homeday.atlassian.net/jira/software/c/projects/REAL/boards/7?modal=detail&selectedIssue=REAL-4538) to work.

### DIFF
--- a/src/components/details-table/HdDetailsTable.vue
+++ b/src/components/details-table/HdDetailsTable.vue
@@ -26,7 +26,8 @@ export default {
     },
     reference: {
       type: String,
-      default: '',
+      required: false,
+      default: undefined,
     },
   },
 };
--- a/src/components/details-table/HdDetailsTable.vue
+++ b/src/components/details-table/HdDetailsTable.vue
@@ -24,6 +24,10 @@ export default {
       type: Boolean,
       default: true,
     },
+    reference: {
+      type: String,
+      default: '',
+    },
   },
 };
 </script>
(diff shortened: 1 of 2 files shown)


volcanioo (approved):

aym3nb (approved):

leandroinacio:
I'm a bit without scope, but how exactly does this new prop affect the component?
Could you add a story on storybook to show why we have it and how to use it?
(don't take this as a blocker to merge though)

leandroinacio:
I'm asking because looks like you just want to bind a reference="x" to this component (based on just looking at the code).
Yet, [inheritAttrs](https://vuejs.org/guide/components/attrs.html#disabling-attribute-inheritance) is set to true here, so you could just pass the attribute and it would stick with the component.

And if someone else use this component now (with this change), we'll always have a reference="", increasing our DOM. 
It would be a problem for us at the website project. (although we are not using, I'm just raising this point in case we ever use)

leandroinacio (approved):

REX-Alexander:
@leandroinacio - I've tried changing things as you suggested in Storybook but it's my first time changing things there so it might all be wrong 😅

leandroinacio (approved):
@REX-Alexander it's not wrong at all! Looks perfect, you we can check the result here:
https://frontend-pulls.homeday.de/homeday-blocks/1007/?path=/docs/components-hddetailstable--default

At the bottom, in the "Docs" tab we can see your description to the prop.
Nice work, and thank you again for taking care of this legacy doc structure 👏 

And fyi for the other reviewers, he had to keep the prop because createElement does not render attributes that are not listed in the Element interface. But as an optional nullable field, shouldn't affect other places unless the prop is explicitly declared.

PR #1012, opened 2022-08-17
✨ Add "uncheck" prop to HdRadio

limhenry:
TBA

### DIFF
--- a/src/components/form/HdRadio.vue
+++ b/src/components/form/HdRadio.vue
@@ -109,6 +109,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    uncheck: {
+      type: Boolean,
+      default: false,
+    },
   },
   data() {
     return {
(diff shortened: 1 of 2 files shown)


SusCasasola (commented):
The code looks good, but it makes me wonder if we should uncheck the radio buttons as we are modifying their default behaviour, shouldn't we use a checkbox instead? or may be a "none of the above" options? I am curious about the specific cases where we could use this

limhenry:
@SusCasasola 

- checkbox: no? because user can select one option only
- "none of the above": um ... sounds like a possible solution? lemme check with the designer
- the biggest problem is this form (sliders) is "optional", so user can proceed without select any option

image

SusCasasola:
> * the biggest problem is this form (sliders) is "optional", so user can proceed without select any option

Aah I see! thanks for the quick answer. If the problem is that the selection should be optional, then we can suggest using a "none of the above" option. 

I wonder what do the rest of the team thinks about it? @homeday-de/frontend-engineering 

I have never asked myself this, so I searched for information and here is what I found: https://ux.stackexchange.com/questions/13511/why-is-it-impossible-to-deselect-html-radio-inputs

PR #1015, opened 2022-08-29
💄 update HdTabsMenu alignment options

petterHD:


logycode (approved):

SusCasasola (approved):

src/components/HdTabsMenu.vue
```diff
@@ -54,13 +54,18 @@ export default {
       type: Boolean,
       default: false,
     },
+    // Available options: left, right, center
+    alignment: {
+      type: String,
+      default: 'left',
```
aym3nb, line 60: I suggest replacing the comment with a validator instead.
```suggestion
    alignment: {
      type: String,
      default: 'left',
      validator: (value) => ['left', 'right', 'center'].includes(value),
```

petterHD: good point! @aym3nb

PR #1025, opened 2022-09-08
:wrench: Add pre-release configuration

alieslamifard:
In order to test new features, we better have a pre-release setup. It can be achieved by publishing Beta/Alpha/RC versions. Since the beta versions can do the job, I set up the configuration to support beta versions only.

Read more: https://github.com/semantic-release/semantic-release/blob/7b7728c33c65ba801cd3bc6162cef4871b4ea0f5/docs/usage/configuration.md#branches

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,7 @@ on:
   push:
     branches:
       - master
+      - beta
 
 jobs:
   release-and-publish:
@@ -26,7 +27,7 @@ jobs:
 
       - name: Install dependencies ⚙️
         run: npm ci
-      
+
       - name: Compile TypeScript
         run: npm run build:tsc
 
--- a/.releaserc.js
+++ b/.releaserc.js
@@ -14,6 +14,7 @@ const commitTemplate = readFileSync(
 ).toString();
 
 module.exports = {
+  branches: ['master', { name: 'beta', prerelease: true }],
   plugins: [
     [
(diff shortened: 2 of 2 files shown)


SusCasasola:
🚀 Thanks for doing this! it's indeed a needed change. Can you also please add on the Readme that we have this option available?

leandroinacio (approved):

aym3nb (approved):
Nice 👏 🚀 

While this is a good option for testing out some bigger changes, I think small changes should still be tested out just by specifying the feature branch in the main project which does work exactly like a beta version.

```json
  "homeday-blocks": "homeday-de/homeday-blocks#new-cool-feature"
```

alieslamifard:
> 🚀 Thanks for doing this! it's indeed a needed change. Can you also please add on the Readme that we have this option available?

Sure, I will add more docs in another PR. Thanks for the suggestion btw.

alieslamifard:
> Nice 👏 🚀
> 
> While this is a good option for testing out some bigger changes, I think small changes should still be tested out just by specifying the feature branch in the main project which does work exactly like a beta version.
> 
> ```json
>   "homeday-blocks": "homeday-de/homeday-blocks#new-cool-feature"
> ```

True, it also worked as we did before by directly installing a branch in the target project, but it looks old-fashioned and not professional to me. Moreover, by publishing pre-releases, we can keep the track of what we did before in other beta releases while working on a future major release.

PR #1029, opened 2022-09-12
⬆️ Add prettier

alieslamifard:
Prettier has been added.

What changes have been made in this PR?
`pakcage.json`
`.eslintrc.js`
`.prettierrc`

Other files have been prettified during the build process according to the [vue-cli-service config](https://github.com/homeday-de/homeday-blocks/blob/master/vue.config.js#L21).


leandroinacio:
Looks good to me, but I think there is a history here about implementing prettier that I don't have full context. I just know it was already debated. Maybe @aym3nb or @drapisarda can help us?

This is all info I have: https://github.com/homeday-de/homeday-blocks/pull/254#issuecomment-565530674

alieslamifard:
> 

Thank you @leandroinacio I guess it is just missed but I will wait for others to debate the need for prettier in the current setup.

leandroinacio (approved):

aym3nb (approved):
Thanks for adding it, totally makes sense since we have it in the main projects too now which was the reason of the debate back then.

PR #1043, opened 2022-09-16
♻️ Descoping styles

drapisarda:
[CA-4193](https://homeday.atlassian.net/browse/CA-4193)

Descoping the styles to reduce the HTML produced.
This would need a beta version release to be then tested on our products


src/components/HdIcon.vue
```diff
@@ -102,3 +102,5 @@ export default {
   },
 };
 </script>
+
+<style lang="scss" scoped></style>
```
leandroinacio, line 106:
```suggestion
<style lang="scss"></style>
```

PR #1044, opened 2022-09-16
:wrench: Add src folder and barrel module to the final bundle

alieslamifard:
### To make the SSR-ready bundle available we added raw components to the final build.

How to use:
In SSR/SSG projects you can directly import components and services from `homeday-blocks/src`.

`import { mediaMatches, HdIcon } from 'homeday-blocks/src';`


leandroinacio (approved):
Nice 🚀 
I suggest adding that part of how to import for SSR/SSG in the readme.md

alieslamifard:
Sure, I have to also add some documents about using beta branch in another PR. @leandroinacio

PR #1049, opened 2022-09-23
♻️ removed annoing warnings

drapisarda:


drapisarda:
> If we remove the warnings we might as well remove the old variables and typography, right? Otherwise no one ever will without the warnings 😅

done

PR #1074, opened 2022-10-13
♻️ Migrate from node-sass to dart-sass

aym3nb:
## Details
As it's written in official [node-sass github doc](https://github.com/sass/node-sass#node-sass), `node-sass` is deprecated.

> Warning: LibSass and Node Sass are deprecated. While they will continue to receive maintenance releases indefinitely, there are no plans to add additional features or compatibility with any new CSS or Sass features. Projects that still use it should move onto Dart Sass.

## Related tickets
[FET-251](https://homeday.atlassian.net/browse/FET-251)

## Main changes
- Remove `node-sass` and add `sass` ([Dart Sass](https://npmjs.com/package/sass))
- Use `calc()` instead of a slash (/) for division because it is deprecated and will be removed in Dart Sass 2.0.0


leandroinacio (approved):
LGTM 👍 
I believe we should pin the new dependency version though.

aym3nb:
> I believe we should pin the new dependency version though.

I don't think it is necessary, `node-sass` version was pinned because newer versions were breaking SB.
But `sass` patch and minor versions should still be safe to use.

PR #1075, opened 2022-10-17
✅ add test to check if boolean values are supported

logycode:
## Ticket

[[FE] Implement support for boolean values in HDRadio component](https://homeday.atlassian.net/browse/DS-209)

## Description

Added unit test to HDRadio.spec to check on boolean values so that we can be sure that they can be used reliably.

### Changes

- HDRadio.spec.js
- FORM_ITEMS.js

### DIFF
--- a/src/stories/mocks/FORM_ITEMS.js
+++ b/src/stories/mocks/FORM_ITEMS.js
@@ -19,4 +19,8 @@ export default [
     label: 'Cersei',
     value: 'cersei',
   },
+  {
+    label: 'Is Batman?',
+    value: true,
+  },
 ];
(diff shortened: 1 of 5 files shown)


aym3nb:
Hey, @logycode `HdRadio` component boolean type support was already implemented a while ago (see [commit](https://github.com/homeday-de/homeday-blocks/pull/954/commits/49edfc270365f247823acd63c25013dd234c93df)) so this ticket is already done.

And this test isn't needed because even if `HdRadio` component doesn't accept boolean values it'll still pass because Vue just emits a warning in that case.

Screen Shot 2022-10-17 at 10 19 23

PR #1095, opened 2022-11-24
✨ New pager modifier

leandroinacio:
**Task**
 https://homeday.atlassian.net/browse/TOFU-4906

**What**
A new layout option for the pager.
Component documentation: https://www.figma.com/file/KdznSjw9btqHEKdgmm2UT4co/Design-System-UI-Patterns?node-id=7%3A100&t=6pwgJeQW0pioLIYG-0

**Why**
This will be used in a new component of the website.
Theresa requested to be reusable, because she will also be using this in other places in the future.
Design: https://www.figma.com/file/0uSB2ZGAOQauErdTfVnZEq/Buyer-Value-component---PA?node-id=187%3A8317&t=spNVZE4rI0yESA4c-0

**Notes**
Layout and animation wise, Theresa already tested and approved.
Now let's see code wise with y'all :)

### DIFF
--- a/src/components/HdPager.vue
+++ b/src/components/HdPager.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     v-if="count > 1"
-    :class="customClasses"
+    :class="customMainDivClasses"
     @keydown="setUsingMouse(false)"
     @mousedown="setUsingMouse(true)"
   >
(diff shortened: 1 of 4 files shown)


src/components/HdPager.vue
```diff
@@ -27,6 +19,7 @@
         class="pager__items__track"
       >
         <div
+          v-if="modifier === 'wide'"
```
aym3nb, line 22: Would be nice if this was a computed prop that makes use of the modifier enum `HdPagerModifierEnum.WIDE.

```suggestion
          v-if="isWide"
```

REX-Alexander (approved):

src/components/HdPager.vue
leandroinacio: Good point, pushed it in https://github.com/homeday-de/homeday-blocks/pull/1095/commits/948912baf2fd53e17d11e791881137a4b320e485

PR #1099, opened 2022-12-06
:bug: Fix sass lexical error

leandroinacio:
Same as https://github.com/homeday-de/homeday-blocks/commit/cddd3751b466f6d0d32c278f19c8e3de5ce1d847
We're facing this same issue on the website after the last update.
image

### DIFF
--- a/src/components/form/HdRange.vue
+++ b/src/components/form/HdRange.vue
@@ -322,11 +322,11 @@ export default {
     &:after {
       content: '';
       display: block;
-      margin-left: calc(-1 * $steps-size / 2);
-      margin-top: calc(-1 * $steps-size / 2);
-      height: $steps-size - ($steps-border-size * 2);
-      width: $steps-size - ($steps-border-size * 2);
-      border-radius: calc($steps-size / 2);
+      margin-left: calc(-1 * #{$steps-size / 2});
+      margin-top: calc(-1 * #{$steps-size / 2});
+      height: #{$steps-size} - #{($steps-border-size * 2)};
+      width: #{$steps-size} - #{($steps-border-size * 2)};
+      border-radius: calc(#{$steps-size} / 2);
       background: $primary-bg;
       border: $steps-border-size solid $quaternary-color;
       border-radius: 50%;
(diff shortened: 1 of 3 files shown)


src/components/form/HdRange.vue
```diff
@@ -322,11 +322,11 @@ export default {
...
       display: block;
-      margin-left: calc(-1 * $steps-size / 2);
-      margin-top: calc(-1 * $steps-size / 2);
-      height: $steps-size - ($steps-border-size * 2);
-      width: $steps-size - ($steps-border-size * 2);
-      border-radius: calc($steps-size / 2);
+      margin-left: calc(-1 * #{$steps-size / 2});
+      margin-top: calc(-1 * #{$steps-size / 2});
+      height: #{$steps-size} - #{($steps-border-size * 2)};
+      width: #{$steps-size} - #{($steps-border-size * 2)};
```
aym3nb, line 328: Is this necessary? because they aren't within a `calc()` so it shouldn't cause a lexical error.
```suggestion
      height: #{$steps-size} - #{($steps-border-size * 2)};
      width: #{$steps-size} - #{($steps-border-size * 2)};
```

aym3nb (approved):
Weird, when I upgraded blocks in myHD the only lexical error occurred in https://github.com/homeday-de/homeday-blocks/commit/cddd3751b466f6d0d32c278f19c8e3de5ce1d847 🤔

src/components/form/HdRange.vue
leandroinacio: Ahh, probably not. I was probably caught up in the ctrl+c/ctrl+v frenzy.
When I upgrade and install, the only error I get is on HdRange.
But the only erroneous area is 
```
1: $number / $divisor
```
So I suspect it's `homeday-blocks/src/styles/mixins.scss`, strip-unit function.
The other calcs, I changed just for the sake of changing, I mean... the syntax should be the same, right?
I'm not fully confident about the other calcs, let me know your thoughts.
If you think we should fix it only where the lexical error was given, I can revert the last commit (and along will go these lines without calc()).

leandroinacio: Alright, thanks for the fast reply.
I've removed the change from this specific part but kept the other ones.

PR #1115, opened 2023-01-12
✨ Add actions menu option to HdToggle

petterHD:
Add option for actions menu.

Multiple actions:
PixelSnap 2023-01-13 at 10 10 03@2x

Single action:
PixelSnap 2023-01-13 at 10 10 44@2x

### DIFF
--- a/src/components/HdToggle.vue
+++ b/src/components/HdToggle.vue
@@ -137,7 +137,7 @@ export default {
     },
     onClickActionsMenu() {
       if (this.actions.length === 1) {
-        this.executeAction(this.actions[0].name);
+        this.$emit(this.actions[0].name);
       } else {
         this.toggleActionsMenu();
       }
--- a/src/components/HdToggle.vue
+++ b/src/components/HdToggle.vue
@@ -50,7 +50,7 @@ import OnResizeService from 'homeday-blocks/src/services/on-resize';
 import HdIcon from 'homeday-blocks/src/components/HdIcon.vue';
 import HdButton from 'homeday-blocks/src/components/buttons/HdButton.vue';
 import { chevron as chevronIcon } from 'homeday-assets';
-import kebabMenuIcon from '@/assets/kebab-menu.svg';
+import { kebabMenu as kebabMenuIcon } from 'homeday-assets';
 
 export const TABINDEX_BACKUP_ATTRIBUTE = 'data-hd-tabindex-backup';
 
--- a/src/components/HdToggle.vue
+++ b/src/components/HdToggle.vue
@@ -13,7 +13,7 @@
       <div class="hd-toggle__control-icon-wrapper">
         <ul
           class="hd-toggle__control-actions-menu"
-          :class="{ 'hd-toggle__control-actions-menu--is-open': actionsMenuIsOpen }"
+          :class="{ 'hd-toggle__control-actions-menu--is-open': isActionsMenuOpen }"
           v-if="actions.length > 1"
         >
           <li v-for="action in actions" :key="action.name" @click.stop="executeAction(action.name)">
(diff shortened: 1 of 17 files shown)


src/components/HdToggle.vue
```diff
@@ -1,17 +1,36 @@
...
     @keydown="setUsingMouse(false)"
     @mousedown="setUsingMouse(true)"
   >
     <button :disabled="!canBeToggled" class="hd-toggle__control" type="button" @click="toggleOpen">
       {{ title }}
-
-      <HdIcon :src="chevronIcon" class="hd-toggle__control-icon" />
+      <div class="hd-toggle__control-icon-wrapper">
+        <ul
+          v-if="actions.length && !hasSignleAction"
```
leandroinacio, line 15:
```suggestion
          v-if="actions.length && !hasSingleAction"
```

src/components/HdToggle.vue
```diff
@@ -64,9 +96,17 @@ export default {
       isUsingMouse: false,
       internalFocusRemoved: false,
       chevronIcon,
+      kebabMenuIcon,
+      isActionsMenuOpen: false,
     };
   },
   computed: {
+    hasSignleAction() {
```
leandroinacio, line 104:
```suggestion
    hasSingleAction() {
```

src/components/HdToggle.vue
```diff
@@ -29,13 +48,16 @@
 <script>
 import OnResizeService from 'homeday-blocks/src/services/on-resize';
 import HdIcon from 'homeday-blocks/src/components/HdIcon.vue';
+import HdButton from 'homeday-blocks/src/components/buttons/HdButton.vue';
 import { chevron as chevronIcon } from 'homeday-assets';
+import kebabMenuIcon from '@/assets/kebab-menu.svg';
```
leandroinacio, line 53: Don't we want to add this to hd-assets?
If we ever reuse it, would be very helpful to have it there.

src/components/HdToggle.vue
```diff
@@ -103,6 +143,20 @@ export default {
     getScrollHeight(el) {
       return el.scrollHeight;
     },
+    toggleActionsMenu() {
+      this.isActionsMenuOpen = !this.isActionsMenuOpen;
+    },
+    onClickActionsMenu() {
+      if (this.hasSignleAction) {
```
leandroinacio, line 150:
```suggestion
      if (this.hasSingleAction) {
```

tests/unit/components/HdToggle.spec.js
```diff
@@ -59,6 +61,29 @@ describe('HdToggle', () => {
...
+          name: 'edit',
+          label: 'Edit',
+          icon: TEST_SVG,
+        },
+      ],
+    });
+
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.html()).toMatchSnapshot();
```
aym3nb, line 84:
```suggestion
```
`HdToggle` already has a snapshot test, so it would be better to test only the actions behavior here:
- Check for the existence of the actions provided
- Check that when an action is clicked it emits an event with its name

src/components/HdToggle.vue
aym3nb:
```suggestion
import { kebabMenu as kebabMenuIcon } from 'homeday-assets'
```
@petterHD a new `homeday-assets` version was published and it has this new kebab menu icon.
Could you please update `homeday-assets` to version `4.56.0`?

aym3nb (changes requested):

tests/unit/components/HdToggle.spec.js
petterHD: Got it, I have re-written the HdToggle test 👍

leandroinacio (approved):

PR #1118, opened 2023-01-16
🏷️  add typescript to HdIcon

petterHD:
add typescript to HdIcon


leandroinacio (changes requested):
We shouldn't have changes on package-lock.json
Check your node/npm version, it should match what we have in .nvmrc

PR #1184, opened 2023-08-23
🐛 Fix HdToast event listeners

alieslamifard:
## Overview
Prevent memory leaks by clearing event listeners before destroying the component.
Let the parent component close the toast by emitting a custom event: `close`.
Rename `close` to `afterClose` custom event, following naming convention best practices.

## Related tickets
- None

## Main changes
- Modified HdToast component to clear event listeners and rename an event.
- Modified stories to add a close button
- Modified unit test

### Before
Screen Shot 1402-06-01 at 12 28 21

### After 
Screen Shot 1402-06-01 at 12 27 45


leandroinacio (commented):
Wow very nice catch!
I would add a line or two to test the emitted close event (in your current change you are testing only the afterClose).

leandroinacio (approved):

aym3nb (approved):

PR #1193, opened 2023-09-15
:sparkles: Return if all fields are valid inside form

SusCasasola:
- Return true / false to know if all the fields of a form are valid or not (either JS or HTML validations).
- Show a custom error message when  HTML validation fails.


src/components/form/HdForm.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <form class="form" @submit.prevent="submit" novalidate>
+  <form class="form" @submit.prevent="submit" novalidate ref="form">
```
aym3nb, line 2:
```suggestion
  <form ref="form" class="form" @submit.prevent="submit" novalidate>
```

aym3nb (approved):

PR #1197, opened 2023-09-20
✨ New Components Structure

SusCasasola:
https://homeday.atlassian.net/browse/DS-242

* I modified the Story titles to match the new Folder structure. I did not move all the files because it would have been more chaotic 😅 but feel free to do it in the future. 

* I am using MDX **only** on the Welcome story and inside the Guidelines folder. (See yellow icons on screenshots). I think it's easier to use MDX format to write stories that are only meant to serve as documentation, with a couple of code examples. When we are importing a Vue component, then is okay to use the CSF.

* I moved all the styles that are used **only** on the stories to the `stories.scss` file. This file is meant to be used only on the `preview.js` file.

| Before | After |
|--------|-------|
| image 1 | image 4 |


leandroinacio (approved):
Amazing!
Docs are looking much better now.
Thank you very much Sussie!

aym3nb (approved):
Well done 👏 🚀

PR #1219, opened 2023-10-30
:sparkles: Adds new tag list modifier

leandroinacio:
https://homeday.atlassian.net/browse/TOFU-5523

The new property carousel requires usage of a different type of tag. 
Since we already have the component, I have added a modifier to reuse the code.

localhost_6006_iframe html_args=modifier_secondary id=components-content-hdtagslist--default viewMode=story

### DIFF
--- a/src/components/HdTagsList.vue
+++ b/src/components/HdTagsList.vue
@@ -1,5 +1,5 @@
 <template>
-  <section class="tags-list">
+  <section class="tags-list" :class="customClasses">
     <div v-for="item in items" :key="item" class="tags-list__tag">
       {{ item }}
     </div>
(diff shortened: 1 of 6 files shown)


src/components/HdTagsList.vue
```diff
@@ -23,18 +32,36 @@ export default {
...
+    #{$root}__tag {
+      height: $sp-l;
+      background-color: getShade($quaternary-color, 50);
+      border-radius: 3px;
+    }
+  }
+  &--secondary {
+    #{$root}__tag {
+      height: $sp-m;
+      background-color: $cello-blue;
```
aym3nb, line 58: [abstraction](https://github.com/homeday-de/homeday-blocks/blob/master/src/styles/_colors.scss#L26)
```suggestion
      background-color: $primary-color;
```

src/components/HdTagsList.vue
```diff
@@ -23,18 +32,36 @@ export default {
...
+      border-radius: 12px;
+      color: $white;
+      font-weight: 600;
+      font-size: 10px;
+      line-height: 17px;
```
aym3nb, line 63: This is probably meant to be `18px` otherwise the text won't be vertically centered.
```suggestion
      line-height: 18px;
```

PR #1247, opened 2024-02-12
:bug: Enables landline phones in HdPhoneInput

leandroinacio:
Not sure if anyone remembers the day we had an issue with the landline phone numbers on the website:
https://homeday.slack.com/archives/C01S70CGKA8/p1670250340393059

> Homeday
030 220 440 000

This was being deemed invalid by the input and blocking some requests.
So, there are plans now of using this input in sliders, so we need to make sure it works for landlines too.

I added a new boolean prop allow or not only mobile phones.
I also fixed the error for duplicated keys that the component had and removed some unnecessary prop binding in the html (which resulted in new snapshots).


aym3nb (approved):
👏🚀

